PR #17, opened 2022-09-08
NathanJumaahBio

njumaah:
Added Nathan Jumaah's bio onto the README


zaverbac (commented):
Having people who want to not just do the bare minimum but also want to work hard is always welcomed.

zaverbac (approved):

PR #59, opened 2022-10-10
fix(client): 🐛 use picsum instead of cataas

nstringham:
the cataas api keeps going down which breaks stuff


njumaah (approved):
page loads faster now and no longer times out. good

njumaah:
Resolve issue with Cataas not loading the image by using Picsum instead.

Issue #26, opened 2022-09-27
splash screen


nstringham:
Catniss should enter a name before joining or creating a lobby because they want to name themselves instead of random names in the game.

- [x] Display Join and Create Lobby Buttons
- [x] Display Input for Display Name
- [x] Account Dropdown

Acceptance Criteria
- Ability to use buttons to head to different pages
- Ability to enter a display name that will be used later
- If the user tries to join a game without a name they will be asked to provide a name
- Ability to access the drop down that will display different options (account, stats, sign out)
- Validate user doc

nstringham:
~~Should probably add material svelte as a dependency as part of this card~~

nstringham:
the current implementation causes way to many database writes

gchamb:
> the current implementation causes way to many database writes

Are you referring to when the display name is changing it is making updates to their firestore doc?

nstringham:
What does "Validate user attributes" mean?

PR #61, opened 2022-10-10
fix(client): 🐛 Correct excessive writes to firestore

gchamb:
- fix excessive writes to firestore by switching on change event to on blur.
- removed user logic from buttons
- renamed function and return type

#26


client/src/lib/firebase/splash.ts
```diff
@@ -6,9 +6,9 @@ import { doc, updateDoc, getDoc, setDoc } from "firebase/firestore";
 //  - displayName
 //  - avatar
 // get display name function to autofill the name input text
-export const getDisplayName = async (id: string) => {
+export const getUser = async (id: string) => {
   const user = await getDoc(doc(db, `/users/${id}`));
-  return user.data() as UserData;
+  return user.data() as UserData | undefined;
```
nstringham: Isn't `UserData | undefined` already the return type of data()?

client/src/routes/+page.svelte
```diff
@@ -1,45 +1,74 @@
...
+  3. A User with User Doc
+    - just update their display name
+  */
+  const saveOrCreate = async () => {
+    if (name === "") {
+      return;
+    }
+    // this is an anon user
+    // create anon user and user doc with display name
+    if (user === null && userData === undefined) {
```
nstringham, line 53: What if the user has logged in after the onmount was called?

client/src/routes/+page.svelte
```diff
@@ -85,10 +114,10 @@
 <main class="cat-main-container">
   <div class="cat-main">
     <div class="logo-container">
-      <img src="https://cataas.com/cat/says/hello%20world!" alt="our logo" />
+      <img src="" alt="our logo" />
     </div>
     <div class="cat-main-buttons">
-      <input type="text" placeholder="Enter in your display name" on:change={saveName} bind:value={name} />
+      <input type="text" placeholder="Enter in your display name" on:blur={saveOrCreate} bind:value={name} />
```
nstringham: Might also consider `oninput` not sure which is better

nstringham (changes requested):
Could also consider creating the user document when the use clicks the create/join lobby button

firestore.rules
```diff
@@ -8,19 +8,17 @@ service cloud.firestore {
...
+        int(request.resource.data.avatar) is number && request.resource.data.avatar >= 0 && request.resource.data.avatar <=12
         );
       }
-        // // only allows the itself to write to the document
-        // allow write: if request.auth.uid == userId;
         // // allows anyone to read that information
-        // allow read;
+        allow read;
+        allow create, update: if request.auth.uid == userId && validateAttrs();
+        allow delete: if request.auth.uid == userId;
```
nstringham, line 22: Are we sure we want users to be able to delete this?

firestore.rules
```diff
@@ -8,19 +8,17 @@ service cloud.firestore {
...
        function validateAttrs(){
         return (
-      		  request.resource.data.keys().hasOnly(['displayName', 'avatar']) &&
+      	request.resource.data.keys().hasOnly(['displayName', 'avatar']) &&
         request.resource.data.displayName is string && request.resource.data.displayName.trim().size() >= 3 && request.resource.data.displayName.trim().size() <= 15 &&
-        // TODO: finish display name regex
-        // request.resource.data.displayName.match('/[0-9A-Za-z_-]/')
-        request.resource.data.avatar is number && request.resource.data.avatar >= 0 && request.resource.data.avatar <=12
+        request.resource.data.displayName.matches('[A-Za-z0-9-_]+') &&
+        int(request.resource.data.avatar) is number && request.resource.data.avatar >= 0 && request.resource.data.avatar <=12
```
nstringham: Wouldn't this allow strings because we are casting to int?

client/src/routes/+page.svelte
```diff
@@ -1,49 +1,76 @@
...
-    }
-    // creates user doc for any user or overwrite if it already exist
-    createUser(user.uid, name);
-
-    const code = (await createLobby()) as string;
-
+    // Create User
+    saveOrCreate();
+    // Creat Lobby
+    const code = await createLobby();
```
nstringham, line 75: This seams like a race condition to me between user and game getting created

nstringham (changes requested):

client/src/lib/firebase/app.ts
```diff
@@ -37,6 +38,9 @@ const app = initializeApp(firebaseConfig);
 export const db = getFirestore(app);
 export const auth = getAuth(app);
 
+auth.onAuthStateChanged((user) => {
+  authStore.set(user);
+});
```
nstringham, line 43: will this code be run every time this component is loaded?
would it be better to put it in `$stores/auth`?

client/svelte.config.js
```diff
@@ -11,6 +11,7 @@ const config = {
     adapter: adapter(),
     alias: {
       $components: "./src/components",
+      $stores: "./src/store",
```
nstringham, line 14: it seams like some people the internet use `$lib/stores` instead of `$stores` but either way seams fine to me

firestore.rules
```diff
@@ -8,19 +8,16 @@ service cloud.firestore {
...
        function validateAttrs(){
         return (
-      		  request.resource.data.keys().hasOnly(['displayName', 'avatar']) &&
+      	request.resource.data.keys().hasOnly(['displayName', 'avatar']) &&
         request.resource.data.displayName is string && request.resource.data.displayName.trim().size() >= 3 && request.resource.data.displayName.trim().size() <= 15 &&
-        // TODO: finish display name regex
-        // request.resource.data.displayName.match('/[0-9A-Za-z_-]/')
-        request.resource.data.avatar is number && request.resource.data.avatar >= 0 && request.resource.data.avatar <=12
+        request.resource.data.displayName.matches('[A-Za-z0-9-_]+') &&
+        request.resource.data.avatar is number &&  int(request.resource.data.avatar) == request.resource.data.avatar &&  request.resource.data.avatar >= 0 && request.resource.data.avatar <=12
```
nstringham: normally I would indent this

package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "cat-me-if-you-can",
+  "name": "cat-project",
```
nstringham: why?

firestore.rules
nstringham: lets try to keep lines less than around 120 charicters long.

you can check by placing you're cusor at the end of the line and seeing that VS-Code says "Ln 14, Col 192" at the bottom. this means that you are on line 14 and coloumn 192 which means that the line is longer than 120 charicters

nstringham (changes requested):
oops I made those comments not part of this review. sorry

package-lock.json
gchamb: I have it named as that on my machine so it looks like it named it that in the package-lock.json. Mistake.

nstringham: hmm. maybe we should set a name in package.json to avoid this

client/src/lib/firebase/app.ts
nstringham: this was me being dumb because I thought this was +page.svelte for some reason

client/svelte.config.js
nstringham: we decided we like `$stores`

nstringham (approved):
code looks good.
tested and working as expected.

PR #63, opened 2022-10-13
feat(client): ✨ Add Copy and Share Function

zaverbac:
Add copy and share function when user clicks on copy button, it copies link to clipboard and when user clicks share button, bring up popup to share link with various methods


nstringham (approved):
this looks good as long as it works in safari

gchamb (approved):
works on safari

Issue #30, opened 2022-09-29
Allow Catniss to receive a role


nstringham:
Upon game start, Catniss should receive either "cat" or "catfish" and it should display that role on the page so that she knows which "team" she is a part of and knows what to do during the game.

A hacker should not be able to view whether other players are a catfish because that would potentially ruin the entire lobby

- [x] Create role receiving screen
- [x] Create function to randomly assign roles
- [x] Make sure only the desired number of catfish is assigned in a lobby (make this configurable)

**Acceptance Criteria**

-  Once game starts, user is given a role either cat or catfish
- Displays the role to the user
- User shouldn't be able to see others role
- Hacker should not be able to see others role
- Hacker should not be able to influence who is cat and who is catfish

PR #67, opened 2022-10-14
feat(client): ✨ Appropriate rerouting for game page based on auth

### LINKED ISSUE #29: Allow Catniss to join an existing lobby
Catniss should be able to join a lobby with a given code or link so that she can play with her friends.

- [x] UI for the join lobby page
- [x] Allow the user to enter the code to join the lobby
- [x] Allow joining the lobby with a linky

**Acceptance Criteria**

- If the user clicks on the correctly shared link, they should be able to join the lobby with other players
- If the user correctly enters the shared lobby code, they should also be able to join the lobby
- If the user enters an incorrect lobby code, they should be given an error and cannot join the lobby
- The lobby page should display the number of players currently in the lobby.
- The start game button should set the next game state 
- If anon user, then create ACC and users document and get their display name

gchamb:
- changed the share url based on josh's suggestion
- Componetized the SigninButton
- reroutes users who are anon and arent in this lobby
- implemented the auth store


client/src/lib/firebase/game.ts
```diff
@@ -0,0 +1,12 @@
...
+import type { GameState } from "./firestore-types/lobby";
+
+/*
+  You can change the game state of the current lobby
+*/
+export const changeGameState = async (lobbyId: string, state: GameState) => {
+  await updateDoc(doc(lobbyCollection, lobbyId), {
+    state,
+  });
+};
```
nstringham: I feel like for security reasons we might want this to be on the server but we can fix this as part of [#30](https://github.com/cat-acronym-team/cat-me-if-you-can/issues/30)

client/src/components/Modal.svelte
```diff
@@ -9,10 +9,20 @@
 
   // Props
   export let open: boolean;
-  export let onClosed: () => void;
+  export let closeOnClick = false;
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  export let onClosed = () => {};
 </script>
```
nstringham, line 17: the recomended way to resolve this warning is to place a comment in the function to make it obvious that you are doing nothing on purpose.
```ts
export let onClosed = () => {
  // Do nothing by default
};
```

client/src/lib/firebase/game.ts
```diff
@@ -0,0 +1,12 @@
+import { doc, updateDoc } from "firebase/firestore";
+import { lobbyCollection } from "$lib/firebase/firestore-collections";
+import type { GameState } from "./firestore-types/lobby";
+
+/*
+  You can change the game state of the current lobby
+*/
+export const changeGameState = async (lobbyId: string, state: GameState) => {
```
nstringham: can this be a `export function` instead of a `export const` please

client/src/lib/firebase/join-lobby.ts
```diff
@@ -0,0 +1,31 @@
+import { lobbyCollection } from "./firestore-collections";
+import { doc, updateDoc, getDoc } from "firebase/firestore";
+import type { UserData } from "./firestore-types/users";
+import type { Player } from "./firestore-types/lobby";
+
+export const findAndJoinLobby = async (id: string, allUserInfo: UserData & { uid: string }) => {
```
nstringham: can this be a `export function` instead of a `export const` please

client/src/lib/firebase/join-lobby.ts
```diff
@@ -0,0 +1,31 @@
...
+    uids: [...uids, allUserInfo.uid],
+    players: [
+      ...players,
+      {
+        displayName: allUserInfo.displayName,
+        avatar: allUserInfo.avatar as Player["avatar"],
+        alive: true,
+      },
+    ],
+  });
```
nstringham: if you want this to be on the client you defintily need to write a security rule for this

client/src/routes/join/+page.svelte
```diff
@@ -0,0 +1,104 @@
...
+    const { displayName, avatar } = await getUser($user.uid) as UserData;
+    try {
+      // enter lobby with the user's info
+      await findAndJoinLobby(code, {
+        displayName,
+        avatar,
+        uid: $user.uid,
+      });
+      // go to game page
+      goto(`/game?=${code}`);
```
nstringham: the url querry should be a key value pair. something like `/game?code=${code}`

client/src/routes/join/+page.svelte
```diff
@@ -0,0 +1,104 @@
...
+    status: false,
+    message: "",
+  };
+  let code: string;
+
+  // Checks to see if the join page has the code query paramter
+  onMount(() => {
+    // gets code from url search
+    // the svelte magic with searchparams wasnt working
+    const queryCode = $page.url.search.split("=")[1];
```
nstringham: this will fail if the url is something like `/join?src=share-button&code=abcxyz

client/src/routes/join/+page.svelte
```diff
@@ -0,0 +1,104 @@
...
+    try {
+      // enter lobby with the user's info
+      await findAndJoinLobby(code, {
+        displayName,
+        avatar,
+        uid: $user.uid,
+      });
+      // go to game page
+      goto(`/game?=${code}`);
+    } catch (err: any) {
```
nstringham: try not to use any.

client/src/routes/game/+page.svelte
```diff
@@ -0,0 +1,56 @@
...
+  import type { Unsubscribe } from "firebase/auth";
+
+  let lobbyData: Lobby;
+  let code: string;
+  // in case we want to unsub from lobby
+  let unsubscribeLobby: Unsubscribe;
+
+  onMount(async () => {
+    // gets code from url search
+    // the svelte magic with searchparams wasnt working
```
nstringham, line 19: it probobly wasn't working because of the mistake on line 46 of the join page
```ts
goto(`/game?=${code}`);
```

client/src/routes/game/+page.svelte
```diff
@@ -0,0 +1,56 @@
...
+    code = $page.url.search.split("=")[1];
+    // We want to get the document immediately because if we wait there's a short delay in getting
+    // the new doc from subscribing below. This makes the redirect slow since we want to check if the user isn't in this lobby
+    const lobbyDoc = await getDoc(doc(lobbyCollection, code));
+    if (!lobbyDoc.exists()) {
+      goto("/join");
```
nstringham: it might be nice to pass the code to join so that it can display an error message

client/src/routes/join/+page.svelte
```diff
@@ -0,0 +1,104 @@
...
+</nav>
+<div class="cat-join-container">
+  <h2>Join Lobby!</h2>
+  {#if error.status}
+    <p style="color:red;">{error.message}</p>
+  {/if}
+  <form on:submit|preventDefault={joinLobby}>
+    {#if $user === null}
+      <input type="text" placeholder="Enter in your display name" bind:value={name} />
+    {/if}
```
nstringham: it might be better to have this textbox always visable so that way if I already have an account I can do a last minute name changebefore I join

nstringham (changes requested):
seams to be working pretty good just a few little things to address

client/src/components/Modal.svelte
nstringham: this comment was supposed to be on line 13 and 14

client/src/routes/join/+page.svelte
```diff
@@ -44,11 +69,12 @@
       });
       // go to game page
       goto(`/game?code=${code}`);
-    } catch (err: any) {
+    } catch (err) {
       // if the lobby doesn't exist then error is thrown
+      const myError = err as Error;
       error = {
```
nstringham, line 55: Is this a safe type assertion? I feel like it might be better to use an if.

gchamb: the err parameter in the catch is either unknown or any so you literally have to tell typescript its type to be able to use it.

nstringham: the reason it is `unknown`  is because we actually don't know what it is. it might not be an error. in JavaScript you can throw any value.

here is what we can do:

at runtime check if it is an error using the JavaScript [instanceof operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof)

if it is an error then you can use it's message

if it is not an error then you still want to display something to the user so we can use the [String Constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/String) without the `new` keyword which can convert any value to a string

we could use an if statement like this:

```ts
catch (err) {
  // if the lobby doesn't exist then error is thrown
  let errorMessage: string;
  if (err instanceof Error) {
    errorMessage = err.message;
  } else {
    errorMessage = String(err);
  }
  error = {
    status: true,
    message: errorMessage,
  };
  code = "";
}
```

or we can do it with the [ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) if you want
```ts
catch (err) {
  // if the lobby doesn't exist then error is thrown
  error = {
    status: true,
    message: err instanceof Error ? err.message : String(err),
  };
  code = "";
}
```

gchamb: Okay that makes sense for the unknown type

client/src/routes/join/+page.svelte
```diff
@@ -0,0 +1,128 @@
...
+      error = {
+        status: true,
+        message: errorMessage,
+      };
+      // do this to remove the errorMessage property in state
+      goto("/join", { replaceState: true });
+    }
+    // gets code from url search
+    // the svelte magic with searchparams wasnt working
+    const queryCode = $page.url.search.split("code=")[1];
```
nstringham, line 33: this will not work with a url that looks like `/join?code=abcxyz&src=share-button`

why not use something like this?
```ts
$page.url.searchParams.get("code")
```

client/src/routes/game/+page.svelte
```diff
@@ -0,0 +1,68 @@
...
+
+  let lobbyData: Lobby;
+  let code: string;
+  // in case we want to unsub from lobby
+  let unsubscribeLobby: Unsubscribe;
+
+  onMount(async () => {
+    // gets code from url search
+    // the svelte magic with searchparams wasnt working
+    code = $page.url.search.split("code=")[1];
```
nstringham: this will not work with a url that looks like `/game?code=abcxyz&src=share-button`

why not use something like this?
```ts
$page.url.searchParams.get("code")
```

client/src/routes/join/+page.svelte
nstringham: this will not work with a url that looks like `/join?code=abcxyz&src=share-button`

why not use something like this?
```ts
$page.url.searchParams.get("code")
```

nstringham (changes requested):
if I am already signed in and I click the join button then the display name box is empty instead of showing the Display name I used last time.

client/src/components/Lobby.svelte
```diff
@@ -25,22 +32,24 @@
...
+    <!-- TODO: Probably Display Users with their avatar and name -->
     <div class="lobby" />
     <div class="start">
-      <button id="start-game" on:click={startgame}>Start Game</button>
+      <button
+        id="start-game"
+        on:click={async () => {
+          await startGame({ code, state: "PROMPT" });
+        }}>Start Game</button
+      >
```
nstringham: the await isn't really doing anything here but it's also not hurting anything so whatever

client/src/components/Lobby.svelte
```diff
@@ -1,15 +1,22 @@
...
+  import { functions } from "$lib/firebase/app";
+  import { httpsCallable } from "firebase/functions";
+  // Props
+  export let players: Player[];
+  export let code: string;
+  // better link to share since it's redirecting to this page anyways
+  // Josh's suggestion that I agreed on
+  let url = `${$page.url.origin}/join?code=${code}`;
   let canShare = false;
+  const startGame = httpsCallable<ChangeState, ChangeStateResponse>(functions, "changeState");
```
nstringham: might be a good idea to put this into a seprate file kinda like we do for `firestore-collections.ts` but this is fine for now

functions/src/firestore-types/lobby.ts
```diff
@@ -23,6 +23,8 @@ export type Player = {
 };
 
 export type GameState = "WAIT" | "PROMPT" | "CHAT" | "VOTE" | "END";
+export type ChangeState = { code: string; state: GameState };
+export type ChangeStateResponse = { error?: string };
```
nstringham: this isn't really a "firestore" type maybe it would be a good idea to make a seprate file for firestore functions types?

functions/src/lobby.ts
```diff
@@ -0,0 +1,37 @@
...
+import { UserData } from "./firestore-types/users";
+
+export const changeState = functions.https.onCall((data: ChangeState, context) => {
+  // no auth then you shouldn't be here
+  if (context.auth === undefined) {
+    return { error: "Not Signed In" };
+  }
+  // return
+  return lobbyCollection.doc(data.code).update({ state: data.state });
+});
```
nstringham: definitly need some security here we don't want any logged in player to be able to chnage the sate to anytrhing whenever they feel like

functions/src/lobby.ts
```diff
@@ -0,0 +1,37 @@
...
+  const { players, uids } = lobbyInfo.data() as Lobby;
+  if (uids.includes(data.uid)) {
+    return { error: "You are already in the lobby!" };
+  }
+  // add player
+  return lobby.update({
+    players: [...players, { displayName: data.displayName, avatar: data.avatar as Player["avatar"], alive: true }],
+    uids: [...uids, data.uid],
+  });
+});
```
nstringham: definitly need secrity and validation here as well

client/src/lib/firebase/join-lobby.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (!validLobby.exists()) {
+    throw new Error("Lobby doesn't exist!");
+  }
+  // First, check if they're already in the lobby
+  const { uids } = validLobby.data();
+  if (uids.includes(allUserInfo.uid)) {
+    throw new Error("You are already in the lobby!");
+  }
+  // make request to server
+  const { data } = await httpsCallable<UserData & { uid: string; code: string }, { error?: string }>(
```
nstringham: probobly a good idea to put this into a seprate file kinda like we do for `firestore-collections.ts`

client/src/lib/firebase/join-lobby.ts
```diff
@@ -0,0 +1,28 @@
...
+    functions,
+    "addPlayer"
+  )({ ...allUserInfo, code: id });
+  if (data.error !== undefined) {
+    throw data.error;
```
nstringham: should probobly wrap this in an error

client/src/lib/firebase/splash.ts
```diff
@@ -1,6 +1,36 @@
...
+/*
+  there are three type of situations that could happen
+  1. A Anon User without User Doc 
+    - create anon user and create user doc with display name
+  2. A User without User Doc
+    - create user doc with display name
+  3. A User with User Doc
+    - just update their display name
+*/
+export const saveOrCreate = async (user: User | null, userData: UserData | undefined, name: string) => {
```
nstringham: can we do `export function` instead of `export const` please

nstringham (changes requested):

client/src/lib/firebase/firestore-functions-types.ts
```diff
@@ -0,0 +1,4 @@
+import type { GameState } from "./firestore-types/lobby";
+
+export type ChangeState = { code: string; state: GameState };
+export type ErrorResponse = { error?: string };
```
nstringham: we should probably share types between the client and the server like we did for firestore-types

client/src/lib/firebase/firestore-functions-types.ts
```diff
@@ -0,0 +1,4 @@
+import type { GameState } from "./firestore-types/lobby";
+
+export type ChangeState = { code: string; state: GameState };
```
nstringham: is this used anywhere anymore?

client/src/lib/firebase/firestore-functions.ts
```diff
@@ -0,0 +1,6 @@
+import { functions } from "$lib/firebase/app";
+import { httpsCallable } from "firebase/functions";
+import type { ErrorResponse } from "./firestore-functions-types";
+
+export const startGame = httpsCallable<{ code: string }, ErrorResponse>(functions, "startGame");
+export const addPlayer = httpsCallable<{ code: string }, ErrorResponse>(functions, "addPlayer");
```
nstringham: we should probobly makes types for the imnput data and share the types between the server and the client so that if we want to change it in the future it will force us to change it in both places

functions/src/lobby.ts
```diff
@@ -0,0 +1,56 @@
+import * as functions from "firebase-functions";
+import { lobbyCollection, userCollection } from "./firestore-collections";
+import { Lobby } from "./firestore-types/lobby";
+import { UserData } from "./firestore-types/users";
+
+export const startGame = functions.https.onCall(async (data: { code: string }, context) => {
```
nstringham: we should probobly makes types for the imnput data and share the types between the server and the client so that if we want to change it in the future it will force us to change it in both places

functions/src/lobby.ts
```diff
@@ -0,0 +1,56 @@
...
+  // check if the request is coming from the host of the game
+  const { uids } = lobby.data() as Lobby;
+  if (context.auth.uid !== uids[0]) {
+    return { error: "Not allowed to start game!" };
+  }
+
+  return lobbyCollection.doc(data.code).update({ state: "PROMPT" });
+});
+
+export const addPlayer = functions.https.onCall(async (data: { code: string }, context) => {
```
nstringham: we should probobly makes types for the imnput data and share the types between the server and the client so that if we want to change it in the future it will force us to change it in both places

functions/src/lobby.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Lobby } from "./firestore-types/lobby";
+import { UserData } from "./firestore-types/users";
+
+export const startGame = functions.https.onCall(async (data: { code: string }, context) => {
+  // no auth then you shouldn't be here
+  if (context.auth === undefined) {
+    return { error: "Not Signed In" };
+  }
+  // get lobby doc
+  const lobby = await lobbyCollection.doc(data.code).get();
```
nstringham, line 17: we don't know that data is an object where code is a string so we should do some validation here.

here is how me and nathan are doing it for avatar a2ff3094d4203331f0dea0ee3f5012b46b5ee47a

functions/src/lobby.ts
```diff
@@ -0,0 +1,56 @@
...
+  return lobbyCollection.doc(data.code).update({ state: "PROMPT" });
+});
+
+export const addPlayer = functions.https.onCall(async (data: { code: string }, context) => {
+  // no auth then you shouldn't be here
+  if (context.auth === undefined) {
+    return { error: "Not Signed In" };
+  }
+  // lobby doc
+  const lobby = lobbyCollection.doc(data.code);
```
nstringham, line 40: we don't know that data is an object where code is a string so we should do some validation here.

here is how me and nathan are doing it for avatar a2ff3094d4203331f0dea0ee3f5012b46b5ee47a

functions/src/lobby.ts
nstringham: maybe we should call this joinLobby now that the uid is not in the input data?

nstringham (changes requested):

functions/src/firestore-types/lobby.ts
```diff
@@ -23,7 +23,6 @@ export type Player = {
 };
 
 export type GameState = "WAIT" | "PROMPT" | "CHAT" | "VOTE" | "END";
-
```
nstringham: lets keep this empty line

functions/src/lobby.ts
```diff
@@ -0,0 +1,65 @@
...
+import { Lobby } from "./firestore-types/lobby";
+import { UserData } from "./firestore-types/users";
+
+export const startGame = functions.https.onCall(async (data: LobbyRequest, context) => {
+  // no auth then you shouldn't be here
+  if (context.auth === undefined) {
+    return { error: "Not Signed In" };
+  }
+  // validate code
+  if (codeCheck(data.code) === false) {
```
nstringham: this asumes that data is an object but it might be something else

functions/src/lobby.ts
```diff
@@ -0,0 +1,65 @@
...
+  return lobbyCollection.doc(data.code).update({ state: "PROMPT" });
+});
+
+export const joinLobby = functions.https.onCall(async (data: LobbyRequest, context) => {
+  // no auth then you shouldn't be here
+  if (context.auth === undefined) {
+    return { error: "Not Signed In" };
+  }
+  // validate code
+  if (codeCheck(data.code) === false) {
```
nstringham: this asumes that data is an object but it might be something else

nstringham (changes requested):

nstringham (approved):
just a few small issues so I fixed them myself in the interest of time

PR #68, opened 2022-10-15
fix(client): 🐛 Fix code not generating inside the loop

mrjoshua520:
Fixed code not generating inside the loop that checks if it is valid. Also made createCode no longer async

### DIFF
--- a/client/src/lib/firebase/create-lobby.ts
+++ b/client/src/lib/firebase/create-lobby.ts
@@ -11,7 +11,7 @@ export async function createLobby(): Promise<string> {
   }
 
   for (let index = 0; index < 5; index++) {
-    const code = await createCode(); // Creates lobby code
+    const code = createCode(); // Creates lobby code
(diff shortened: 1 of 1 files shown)


client/src/lib/firebase/create-lobby.ts
```diff
@@ -4,14 +4,14 @@ import { auth } from "./app";
...
 export async function createLobby(): Promise<string> {
-  const code = await createCode(); // Creates lobby code
   let user = auth.currentUser?.uid;
 
   if (user == undefined) {
     user = (await loginAnonymous()).user.uid;
   }
 
   for (let index = 0; index < 5; index++) {
+    const code = await createCode(); // Creates lobby code
```
nstringham: If it's not async you don't need to await it

nstringham (changes requested):

nstringham (approved):

PR #71, opened 2022-10-17
fix(client): 🐛 Fix display name saving as "Default" 

mrjoshua520:
Display name no longer saves as "Default" when creating a lobby. Whatever name entered will be saved instead


njumaah (approved):

nstringham (approved):
seams fine

PR #74, opened 2022-10-17
feat(client): ✨ Add Win/Loss Screen

### LINKED ISSUE #36: Allow catniss to see the win page or loss page
Let catniss see the win or loss page after game ends so that she is able to tell if she won or loss at the end of the game

This also includes removing data from firestore to ensure data is not constantly filing up from new games being created

- [x] Create win screen (for Cats)
- [x] Create loss screen (for Cats)
- [x] Create win screen (for Catfish)
- [x] Create loss screen (for Catfish)
- [x] Create function to detect win or loss after vote
- [x] Set people back to alive upon end

**Acceptance Criteria**
- If cat loss, display loss screen
- If cat win , display win screen
- if catfish win, display win screen
- if catfish loss, display loss screen
- firestore lobby is reset to be read for a new game

SBAsai:
Adds the win/loss screens for the players to view at the end of the game. Different screens are shown depending on the player's role and the game's outcome


nstringham (changes requested):
can you pull from development please

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,168 @@
...
+        }
+      } else {
+        if (aliveCatFishCount == 0) {
+          end = "Catfish Lose";
+        } else if (aliveCatCount == aliveCatFishCount) {
+          end = "Catfish Win";
+        }
+      }
+    }
+  }
```
nstringham, line 48: missing `)`

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,168 @@
...
+
+  let end: string = "";
+  let aliveCount = 0;
+  let aliveCatCount = 0;
+  let catfishDisplayname = "";
+  let currentRole: PrivatePlayer; // the player's current role
+  let lobbyRoles: PrivatePlayer; // cycle through the lobby and retrieve their roles
+
+  function lobbyReturn() {
+      // TODO: change gamestate back to lobby
```
nstringham: we probobly don't want to do that on the client side

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,168 @@
...
+      console.log("Returned to lobby");
+  }
+  onMount (async () => {
+    const lobby = doc(lobbyCollection, code);
+    const validLobby = await getDoc(lobby);
```
nstringham: why not use `export let`?

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,168 @@
...
+  import { onMount } from "svelte";
+
+  export let code: string;
+
+  let end: string = "";
+  let aliveCount = 0;
+  let aliveCatCount = 0;
+  let catfishDisplayname = "";
+  let currentRole: PrivatePlayer; // the player's current role
+  let lobbyRoles: PrivatePlayer; // cycle through the lobby and retrieve their roles
```
nstringham: why is this not scoped to the loop where it is used?

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,168 @@
...
+        aliveCount++;
+        // count the number of remaining players who are alive
+        let privatePlayer = await getDoc(doc(getPrivatePlayerCollection(lobby), uids[i]));
+        lobbyRoles = privatePlayer.data() as PrivatePlayer;
+
+        if (lobbyRoles.role == "CAT") {
+          aliveCatCount++;
+          // within the alive players, count the number that are cats
+        } else {
+          catfishDisplayname = players[i].displayName;
```
nstringham: what if there is more than one catfish

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,168 @@
...
+        }
+      }
+    }
+    let aliveCatFishCount = aliveCount - aliveCatCount;
+    // any other player who is alive and not a cat is a catfish
+    if ($user !== null) {
+      const currentPlayer = await getDoc(doc(getPrivatePlayerCollection(lobby), $user.uid));
+      currentRole = currentPlayer.data() as PrivatePlayer;
```
nstringham: why not use `export let`

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,168 @@
+<script lang="ts">
+  import { doc, getDoc } from "firebase/firestore";
+  import { getPrivatePlayerCollection, lobbyCollection } from "$lib/firebase/firestore-collections";
+  import { authStore as user } from "$stores/auth";
+  import type { PrivatePlayer } from "../../../functions/src/firestore-types/lobby";
+  import { onMount } from "svelte";
+
+  export let code: string;
+
+  let end: string = "";
```
nstringham: the type of this should probobly be enumerated
```ts
let end: "Cat Win" | "Cat Lose" | "Catfish Lose" | "Catfish Win" | null = null;
```
or something like that

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,168 @@
...
+    </div>
+
+    <div class="image" />
+
+    <div class="displayname">
+      <h2>{catfishDisplayname} was sentenced to 9 lives in purrison!</h2>
+      <button on:click={lobbyReturn}>Return to Lobby</button>
+    </div>
+  {/if}
+  {#if (end = "Cat Lose")}
```
nstringham: it would probobly be better to use else if

nstringham (changes requested):
this component needs to be imported and used in the game route

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,164 @@
...
+    <div class="banner">
+      <h2>You have been caught (and released)!</h2>
+      <h2>Might as well have been a <b>clown</b> fish...</h2>
+    </div>
+
+    <div class="image" />
+
+    <div class="displayname">
+      <h2>You should go back to your sea ani-anim-aneme... home</h2>
+      <button on:click={()=>lobbyReturn}>Return to Lobby</button>
```
nstringham, line 118: this does not invoke lobbyReturn

client/src/lib/firebase/winloss.ts
```diff
@@ -0,0 +1,29 @@
...
+  const currentPlayer = await getDoc(doc(getPrivatePlayerCollection(getLobby(code)), id));
+  return currentPlayer.data();
+};
+
+export const lobbyReturn = async (code: string) => {
+  const lobbyData = await getLobbyData(code);
+  if (lobbyData !== undefined) {
+    lobbyData.state = "WAIT";
+  }
+};
```
nstringham, line 29: this doesn't actually do anyhting because the variable `lobbyData` is garbage collected after going out of scope so the state change is lost forever.

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,164 @@
...
+        if (lobbyData.players[i].alive == true) {
+          aliveCount++;
+          // count the number of remaining players who are alive
+          lobbyRoles = await getPrivatePlayer(lobbyCode, lobbyData.uids[i]);
+          if (lobbyRoles !== undefined) {
+            if (lobbyRoles.role == "CAT") {
+              aliveCatCount++;
+              // within the alive players, count the number that are cats
+            } else {
+              catfishDisplayname[i] = lobbyData.players[i].displayName; // create a list of catfishes
```
nstringham, line 33: doesn't this create a array with holes in it?

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,164 @@
...
+            }
+          }
+        }
+      }
+    }
+    let aliveCatFishCount = aliveCount - aliveCatCount;
+    catfishList = [catfishDisplayname.slice(0, -1).join(', '), catfishDisplayname.slice(-1)[0]].join(catfishDisplayname.length < 2 ? '' : ' and ');
```
nstringham: you might want to consider `Intl.ListFormat()` which makes english lists for you https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat/ListFormat

client/src/components/WinLoss.svelte
nstringham: doesn't this create a array with holes in it?

client/src/components/WinLoss.svelte
nstringham: this does not invoke lobbyReturn

client/src/lib/firebase/winloss.ts
nstringham: this doesn't actually do anyhting because the variable `lobbyData` is garbage collected after going out of scope so the state change is lost forever.

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,164 @@
...
+    // any other player who is alive and not a cat is a catfish
+    if ($user !== null) {
+        currentRole = await getCurrentPlayer(lobbyCode, $user.uid); // grab the current user's role
```
nstringham: I really think `export let` would be better here because we already have that data in the game page

nstringham (changes requested):

client/src/lib/firebase/winloss.ts
SBAsai: Would I want to use setDoc here instead to change the game state?

client/src/components/WinLoss.svelte
SBAsai: Does this mean I should make the function under the functions folder or under lib/firebase?

nstringham: it should be in firebase functions aka `/functions/src/` aka server side

client/src/lib/firebase/winloss.ts
nstringham: well yes `setDoc ` would work but it would probably be better to do this on the server in firebase functions

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,163 @@
...
+  let catfishCounter = 0;
+
+  onMount(async () => {
+    let aliveCatCount = 0;
+    // function to count the number of players currently alive
+    if (lobby !== undefined) {
+      for (let i = 0; i < lobby.uids.length; i++) {
+        if (lobby.players[i].alive == true) {
+          // count the number of remaining players who are alive
+          lobbyRoles = await getPrivatePlayer(lobbyCode, lobby.uids[i]);
```
nstringham: you can't read someone elses private player so this code should probobly be on the server side

client/src/lib/firebase/winloss.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const getPrivatePlayer = async (code: string, id: string) => {
+  const privatePlayer = await getDoc(doc(getPrivatePlayerCollection(getLobby(code)), id));
+  return privatePlayer.data();
+};
+
+export const getCurrentPlayer = async (code: string, id: string) => {
+  const currentPlayer = await getDoc(doc(getPrivatePlayerCollection(getLobby(code)), id));
+  return currentPlayer.data();
+};
```
nstringham: this function is identical to getPrivatePlayer

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,163 @@
...
+          if (lobbyRoles !== undefined) {
+            if (lobbyRoles.role == "CAT") {
+              aliveCatCount++;
+              // within the alive players, count the number that are cats
+            } else {
+              catfishDisplayname[catfishCounter] = lobby.players[i].displayName; // create a list of catfishes
+              catfishCounter++;
```
nstringham: did you know about `push`?
what you did works but push will make your life easier if you didn't already know about it.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push

SBAsai:
Lint functions is failing because of import statements, but I cannot figure out why it's throwing an error

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,184 @@
...
+    if (role == "CAT") {
+      if (catfishCount == 0) {
+        return "Cat Win";
+      }
+      if (catCount == catfishCount) {
+        return "Cat Lose";
+      }
+      if (catCount < catfishCount) {
+        return "Cat Lose";
+      }
```
nstringham, line 27: this should proboly be an else case

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,184 @@
...
+    } else {
+      if (catCount == 0) {
+        return "Catfish Win";
+      }
+      if (catfishCount == catCount) {
+        return "Catfish Win";
+      }
+      if (catfishCount == 0) {
+        return "Catfish Lose";
+      }
```
nstringham, line 37: what if there are 2 catfish and 1 cat?

client/src/components/WinLoss.svelte
nstringham: this should proboly be an else case

client/src/components/WinLoss.svelte
nstringham: what if there are 2 catfish and 1 cat?

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,184 @@
+<script lang="ts">
+  import { authStore as user } from "$stores/auth";
+  import type { PrivatePlayer, Lobby } from "../../../functions/src/firestore-types/lobby";
+  import { onMount } from "svelte";
+  import { lobbyReturn, getLobbyRole } from "$lib/firebase/firestore-functions";
+  import type { ErrorResponse, LobbyRoleResponse } from "$lib/firebase/firestore-functions-types";
+
+  export let lobbyCode: string;
+  export let privatePlayer: PrivatePlayer | undefined;
+  export let lobby: Lobby | undefined;
```
nstringham: I don't think these will ever be undefined

functions/src/winloss.ts
```diff
@@ -0,0 +1,68 @@
...
+
+async function getPrivatePlayer(lobbyDocRef: firestore.DocumentReference<Lobby>, id: string) {
+  const privatePlayerCollection = getPrivatePlayerCollection(lobbyDocRef);
+
+  const privatePlayerDoc = privatePlayerCollection.doc(id);
+  const privatePlayer = await privatePlayerDoc.get();
+  return privatePlayer.data();
+}
+
+export const getLobbyRole = functions.https.onCall(async (data: unknown, context) => {
```
nstringham: this should probobly not be run for every client. it should run only once and then the results should be put in the database so everyone can see it

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,184 @@
...
+
+    <div class="displayname">
+      {#if catfishDisplayname.length > 1}
+        <h2>{catfishList} were sentenced to 9 lives in purrison!</h2>
+      {:else}
+        <h2>{catfishList} was sentenced to 9 lives in purrison!</h2>
+      {/if}
+      {#if lobby !== undefined}
+        <button
+          on:click={async () => {
```
nstringham: you don't need async here

nstringham (changes requested):

client/src/components/WinLoss.svelte
SBAsai: I'm doing it the same way start game is doing it so wouldn't the button invoking be the same here too?

nstringham: it's also unnecessary on the start game button thanks for pointing that out

functions/src/winloss.ts
```diff
@@ -0,0 +1,26 @@
...
+  const lobby = await lobbyCollection.doc(data.code).get();
+  if (lobby.exists === false) {
+    return { error: "Lobby doesn't exist!" };
+  }
+
+  const { uids } = lobby.data() as Lobby;
+  if (context.auth.uid !== uids[0]) {
+    return { error: "Not allowed to return to lobby!" };
+  }
+  return lobbyCollection.doc(data.code).update({ state: "WAIT" });
```
nstringham: we should also some code to delete private players and set everyone to alive and maybe some other stuff I am forgeting

functions/src/winloss.ts
```diff
@@ -0,0 +1,26 @@
...
+  // validate code
+  if (!isLobbyRequest(data)) {
+    return { error: "Invalid lobby code!" };
+  }
+  // get lobby doc
+  const lobby = await lobbyCollection.doc(data.code).get();
+  if (lobby.exists === false) {
+    return { error: "Lobby doesn't exist!" };
+  }
+
```
nstringham, line 24: we should probobly also check to see if the state of the game is `END`

functions/src/firestore-types/lobby.ts
```diff
@@ -24,6 +24,11 @@ export type Player = {
    * the number of players that have voted for this player
    */
   votes?: number;
+
+  /**
+   * the role that the player played; used at the end of the game
+   */
+  role: Role;
```
nstringham: this should probobly be optional

functions/src/lobby.ts
```diff
@@ -185,3 +193,74 @@ export const collectPromptAnswers = functions.firestore
...
+    // private player info
+    const privateInfo = privatePlayers[i].data();
+    // get player
+    const player = players[uids.indexOf(privateId)];
+    // replace the role of the player
+    player.role = privateInfo.role;
+  }
+  let catCount = 0;
+  let catfishCount = 0;
+  const catfishDisplayname: string[] = [];
```
nstringham: looks like `catfishDisplayname` is never used

functions/src/lobby.ts
```diff
@@ -185,3 +193,74 @@ export const collectPromptAnswers = functions.firestore
...
+    else {
+      if (players[i].role == "CATFISH") {
+        let temp = "";
+        temp = players[i].displayName;
+        catfishDisplayname.push(temp);
+      }
+    }
+  }
+
+  let winner: Role | undefined;
```
nstringham: this should probobly just be `Role`

functions/src/lobby.ts
```diff
@@ -97,6 +97,14 @@ export const onLobbyUpdate = functions.firestore.document("/lobbies/{code}").onU
   if (lobby.state == "PROMPT" && oldLobby.state != "PROMPT") {
     await startPrompt(lobbyDocRef);
   }
+
+  if (lobby.state == "END" && oldLobby.state != "END") {
+    addRole(lobbyDocRef);
+  }
```
nstringham: what sets the state to `END`?

functions/src/lobby.ts
```diff
@@ -185,3 +193,74 @@ export const collectPromptAnswers = functions.firestore
...
+    }
+  }
+
+  let winner: Role | undefined;
+  if (catfishCount == 0) {
+    winner = "CAT";
+  }
+  if (catCount <= catfishCount) {
+    winner = "CATFISH";
+  }
```
nstringham: these do not cover every case

client/src/lib/firebase/firestore-functions.ts
```diff
@@ -1,8 +1,10 @@
...
 import { httpsCallable } from "firebase/functions";
-import type { LobbyRequest, ErrorResponse } from "./firestore-functions-types";
+import type { LobbyRequest, ErrorResponse, LobbyRoleResponse } from "./firestore-functions-types";
 import type { ChangeAvatarData } from "./functions-types/avatar";
 
 export const startGame = httpsCallable<LobbyRequest, ErrorResponse>(functions, "startGame");
 export const joinLobby = httpsCallable<LobbyRequest, ErrorResponse>(functions, "joinLobby");
+export const lobbyReturn = httpsCallable<LobbyRequest, ErrorResponse>(functions, "lobbyReturn");
 export const changeAvatar = httpsCallable<ChangeAvatarData, void>(functions, "changeAvatar");
+export const getLobbyRole = httpsCallable<LobbyRequest, LobbyRoleResponse | ErrorResponse>(functions, "getLobbyRole");
```
nstringham: looks like this doesn't exist

functions/src/firestore-functions-types.ts
```diff
@@ -6,3 +6,8 @@ export function isLobbyRequest(data: unknown): data is LobbyRequest {
   // will only return true if the data is an object with a code property and string
   return data != null && typeof data === "object" && "code" in data && typeof (data as LobbyRequest).code === "string";
 }
+
+export type LobbyRoleResponse = {
+  aliveCatCount: number;
+  catfishDisplayname: string[];
+};
```
nstringham: looks like this is never used

client/src/routes/game/+page.svelte
```diff
@@ -28,7 +31,7 @@
       return;
     }
 
-    const lobbyDocRef = doc(lobbyCollection, lobbyCode);
+    lobbyDocRef = doc(lobbyCollection, lobbyCode);
```
nstringham: why is this changed from const to let?

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,201 @@
...
+    // get the list of catfishes from lobby.ts and put them into a variable
+    catfishes = lobby.players
+      .filter((p) => {
+        return p.role == "CATFISH";
+      })
+      .map((c) => c.displayName);
+    catfishList = formatter.format(catfishes);
+
+    // grab the current user's role
+    const myRole = (await getDoc(doc(getPrivatePlayerCollection(lobbyDocRef), $user?.uid))).data() as PrivatePlayer;
```
nstringham: this is now on the lobby so we don't need to do an extra database read

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,201 @@
...
+        <h2>{catfishList} were sentenced to 9 lives in purrison!</h2>
+      {:else}
+        <h2>{catfishList} was sentenced to 9 lives in purrison!</h2>
+      {/if}
+      <!-- {#if isHost} -->
+      <button
+        on:click={() => {
+          lobbyReturn({ code: lobbyCode });
+        }}>Return to Lobby</button
+      >
```
nstringham: in `winloss.ts` you say:
```ts
if (context.auth.uid !== uids[0]) {
  return { error: "Not allowed to return to lobby!" };
}
```
if that's how you are going to make it work then this button should only be visable to the host

nstringham (changes requested):

client/src/components/WinLoss.svelte
SBAsai: I forgot to comment it out, but since we're adding timers now, the lobby return button is not going to be used. I just have the lobbyReturn function to test until the timers get added

client/src/routes/game/+page.svelte
SBAsai: I used lobbyDocRef in an export let for my own WinLoss.svelte component so I had to change that

client/src/components/WinLoss.svelte
nstringham: Gotcha.

functions/src/lobby.ts
SBAsai: I'm assuming voting will have a function that will change the lobby state  right?

nstringham: Hmm you and @cmcclusky should talk and make sure you aren't each assuming that eachother is going to do that

SBAsai: For sure, I'll message him

functions/src/lobby.ts
SBAsai: I'm honestly struggling to figure out what case I'm missing. `if(catCount == 0)`?

functions/src/firestore-types/lobby.ts
SBAsai: Sorry can you expand on this? Is there an optional type in Typescript I'm blanking on?

functions/src/lobby.ts
nstringham: For example if there is 1 catfish and 3 cats.

I feel like this will be more obvious once you integrate with the voting code

SBAsai: If there’s 1 catfish and 3 cats the game would continue right? I think neither role wins yet in that situation

mrjoshua520 (dismissed):

functions/src/lobby.ts
```diff
@@ -213,6 +223,89 @@ export const collectPromptAnswers = functions.firestore
...
+
+  let winner: Role;
+  if (catfishCount == 0) {
+    winner = "CAT";
+  } else {
+    // if (catCount <= catfishCount) {
+    winner = "CATFISH";
+  }
+
+  return lobbyDocRef.update({ players, winner });
```
nstringham: this should use a transaction because we are reading data and then writing it back

functions/src/lobby.ts
```diff
@@ -213,6 +223,89 @@ export const collectPromptAnswers = functions.firestore
     });
   });
 
+export async function addRole(lobbyDocRef: firestore.DocumentReference<Lobby>) {
+  const lobby = await lobbyDocRef.get();
+  if (lobby.exists === false) {
+    return { error: "Lobby doesn't exist!" };
```
nstringham: this should be a throw

functions/src/lobby.ts
```diff
@@ -213,6 +223,89 @@ export const collectPromptAnswers = functions.firestore
...
+  }
+
+  return lobbyDocRef.update({ players, winner });
+}
+
+// when returning to the lobby, make every player alive again
+export async function makeAlive(lobbyDocRef: firestore.DocumentReference<Lobby>) {
+  const lobby = await lobbyDocRef.get();
+  if (lobby.exists === false) {
+    return { error: "Lobby doesn't exist!" };
```
nstringham: this should be a throw

functions/src/lobby.ts
```diff
@@ -213,6 +223,89 @@ export const collectPromptAnswers = functions.firestore
...
+  }
+
+  // loop through each player and set their alive status to true
+  const { players } = lobby.data() as Lobby;
+  for (let i = 0; i < players.length; i++) {
+    players[i].alive = true;
```
nstringham: should probobly also remove role from player

functions/src/lobby.ts
```diff
@@ -213,6 +223,89 @@ export const collectPromptAnswers = functions.firestore
...
+  }
+  return lobbyDocRef.update({ players });
```
nstringham: this should also be a transaction

functions/src/lobby.ts
```diff
@@ -213,6 +223,89 @@ export const collectPromptAnswers = functions.firestore
...
+  let deleted = 0;
+
+  const lobby = await lobbyDocRef.get();
+  if (lobby.exists === false) {
+    return { error: "Lobby does not exist!" };
+  }
+
+  // loop through each doc in the private player collection via uid and delete them one-by-one
+  const { uids } = lobby.data() as Lobby;
+  for (let i = 0; i <= docCount; i++) {
```
nstringham: why not use `uids.length` instead of `docCount`

functions/src/lobby.ts
```diff
@@ -213,6 +223,89 @@ export const collectPromptAnswers = functions.firestore
...
+export async function deletePrivatePlayers(
+  lobbyDocRef: firestore.DocumentReference<Lobby>,
+  privatePlayerCollectionRef: firestore.CollectionReference<PrivatePlayer>,
+  docCount: number
+) {
+  let deleted = 0;
+
+  const lobby = await lobbyDocRef.get();
+  if (lobby.exists === false) {
+    return { error: "Lobby does not exist!" };
```
nstringham: this should probobly be a throw

functions/src/lobby.ts
```diff
@@ -213,6 +223,89 @@ export const collectPromptAnswers = functions.firestore
...
+  const { uids } = lobby.data() as Lobby;
+  for (let i = 0; i <= docCount; i++) {
+    const playerDoc = privatePlayerCollectionRef.doc(uids[i]);
+    await playerDoc.delete();
+    deleted++;
+  }
+
+  if (deleted >= docCount) {
+    return;
+  }
```
nstringham: what does this do?

functions/src/lobby.ts
```diff
@@ -213,6 +223,89 @@ export const collectPromptAnswers = functions.firestore
...
+
+  // loop through each player and set their alive status to true
+  const { players } = lobby.data() as Lobby;
+  for (let i = 0; i < players.length; i++) {
+    players[i].alive = true;
+  }
+  return lobbyDocRef.update({ players });
+}
+
+export async function deletePrivatePlayers(
```
nstringham: it might be a good idea to querry instead of using uids from lobby but I'm not sure

functions/src/winloss.ts
```diff
@@ -0,0 +1,27 @@
+import * as functions from "firebase-functions";
+import { lobbyCollection } from "./firestore-collections";
+import { isLobbyRequest } from "./firestore-functions-types";
+import { Lobby } from "./firestore-types/lobby";
+
+// this function will likely be obsolete once timers are introduced
+export const lobbyReturn = functions.https.onCall(async (data: unknown, context) => {
```
nstringham: lets put an explict return type here like 191b0bf80c59858ed7b9e1b52d2e04ce9b9f2f9f

functions/src/winloss.ts
```diff
@@ -0,0 +1,27 @@
...
+  if (context.auth === undefined) {
+    return { error: "Not Signed In" };
```
nstringham: please use errors like in aea453b216d98c6eef8187ee8791a3f67e9acc9f

functions/src/winloss.ts
```diff
@@ -0,0 +1,27 @@
...
+  if (lobby.exists === false) {
+    return { error: "Lobby doesn't exist!" };
+  }
+
+  const { uids } = lobby.data() as Lobby;
+  if (context.auth.uid !== uids[0]) {
+    return { error: "Not allowed to return to lobby!" };
+  }
+  return lobbyCollection.doc(data.code).update({ state: "WAIT" });
+});
```
nstringham, line 38: I feel like the lobby reseting stuff should be done here instead of in `onLobbyUpdate`

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,197 @@
...
+<div class="container">
+  <!-- Check the user's role and display the correct win or loss page accordingly -->
+  {#if end == "Cat Win"}
+    <div class="banner">
+      <h2>Hooray!</h2>
+      <h2>You have sniffed out the cat fish!</h2>
+    </div>
+
+    <div class="image">
+      <img src="https://www.azpetvet.com/wp-content/uploads/Happy-cat-1024x711.jpg" alt="clown fish" />
```
nstringham: do we have rights to this image?

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,197 @@
...
+      {/if} -->
+    </div>
+  {:else if end == "Cat Lose"}
+    <div class="banner">
+      <h2>You have cat to be kitten me!</h2>
+      <h2>The impawster was not caught!</h2>
+    </div>
+
+    <div class="image">
+      <img src="https://pbs.twimg.com/media/EAmr-PAWsAEoiWR.jpg" alt="cat fish" />
```
nstringham: do we have rights to this image?

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,197 @@
...
+        <span class="button">Waiting for Host to Return to Lobby...</span> -->
+    </div>
+  {:else if end == "Catfish Win"}
+    <div class="banner">
+      <h2>O-fish-ally the greatest!</h2>
+      <h2>Way to deceive your furry friends!</h2>
+    </div>
+
+    <div class="image">
+      <img src="https://media.fisheries.noaa.gov/2020-10/blue%20catfish%20face.jpg?VersionId=null" alt="cat fish" />
```
nstringham: do we have rights to this image?

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,197 @@
...
+      <!-- <span class="button">Waiting for Host to Return to Lobby...</span> -->
+    </div>
+  {:else if end == "Catfish Lose"}
+    <div class="banner">
+      <h2>You have been caught (and released)!</h2>
+      <h2>Might as well have been a <b>clown</b> fish...</h2>
+    </div>
+
+    <div class="image">
+      <img src="https://cdn.drawception.com/drawings/577991/0C653P5Wka.png" alt="clown fish" />
```
nstringham: do we have rights to this image?

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,197 @@
...
+        }}>Return to Lobby</button
+      >
+      <!-- {:else}
+        <span class="button">Waiting for Host to Return to Lobby...</span>
+      {/if} -->
+    </div>
+  {:else if end == "Cat Lose"}
+    <div class="banner">
+      <h2>You have cat to be kitten me!</h2>
+      <h2>The impawster was not caught!</h2>
```
nstringham: it would be nice to say who they were

nstringham (changes requested):
css needs some work but I can fix that later

functions/src/winloss.ts
SBAsai: I commented in line 6 that this function probably won't be used in the future, do you want me to just remove it completely now or when the timers are actually implemented and fix it now?

client/src/components/WinLoss.svelte
SBAsai: I'm just kind of using these images as placeholders for this sprint. I'm not entirely sure if we've come to a decision on what we want here yet

SBAsai: I'm just kind of using these images as placeholders for this sprint. I'm not entirely sure if we've come to a decision on what we want here yet

functions/src/winloss.ts
nstringham: that depends on how close timers are to being done

functions/src/lobby.ts
SBAsai: I honestly just took stuff from the documentation that looked important, but I think the original use case was having a check with a recursive call after the return. I don't think it's necessary here though so I'll remove it

functions/src/lobby.ts
SBAsai: If it becomes an issue I'll look further into it

functions/src/winloss.ts
SBAsai: For now I'll just keep the function updated with the new changes

functions/src/winloss.ts
SBAsai: Would I create a function similar to `onLobbyUpdate` or is there a separate way to call the functions in my own .ts file?

client/src/components/WinLoss.svelte
SBAsai: I went ahead and just drew some pictures in paint so let me know how you feel about them

functions/src/winloss.ts
```diff
@@ -0,0 +1,130 @@
...
+  await resetLobby(lobby.ref);
+});
+
+export function findWinner(lobbyDocRef: firestore.DocumentReference<Lobby>) {
+  return db.runTransaction(async (transaction) => {
+    const lobby = await transaction.get(lobbyDocRef);
+    const lobbyData = lobby.data();
+    if (lobby === undefined) {
+      throw new Error("Lobby does not exist!");
+    }
```
nstringham: this isn't required

functions/src/winloss.ts
```diff
@@ -0,0 +1,130 @@
...
+    if (lobbyData === undefined) {
+      throw new Error("Lobby does not exist!");
+    }
+    const privatePlayerCollection = getPrivatePlayerCollection(lobbyDocRef);
+    const { players, uids } = lobby.data() as Lobby;
```
nstringham: this should be `lobbyData` instead of calling `lobby.data()` agian

functions/src/winloss.ts
```diff
@@ -0,0 +1,130 @@
...
+    for (let i = 0; i < uids.length; i++) {
+      // check the alive players
+
+      if (lobbyData.players[i].alive == true) {
+        if (lobbyData.players[i].role == "CATFISH") {
+          catfishCount++;
+        }
+      }
+    }
+    console.log(lobbyData.winner);
```
nstringham: let's not use console.log in functions on production

functions/src/winloss.ts
```diff
@@ -0,0 +1,130 @@
...
+    if (catfishCount == 0) {
+      lobbyData.winner = "CAT";
+    } else {
+      lobbyData.winner = "CATFISH";
+    }
+    console.log(lobbyData.winner);
```
nstringham: let's not use console.log in functions on production

functions/src/winloss.ts
```diff
@@ -0,0 +1,130 @@
...
+    // loop through each player and set their alive status to true and remove the role property from the player
+    let { players } = lobbyData;
+    players = players.map(({ role, ...rest }) => {
+      return { ...rest, alive: true };
+    });
+
+    // delete the private player collection
+    privatePlayerSnapshot.docs.map((privatePlayerDoc) => {
+      privatePlayerDoc.ref.delete();
+    });
```
nstringham: this should probobly be a for loop

functions/src/winloss.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  // loop through each doc in the private player collection via uid and delete them one-by-one
+  const { uids } = lobby.data() as Lobby;
+  for (let i = 0; i <= uids.length; i++) {
+    const playerDoc = privatePlayerCollectionRef.doc(uids[i]);
+    await playerDoc.delete();
+  }
+  return;
+}
```
nstringham: this is unused

nstringham (changes requested):

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,217 @@
...
+      } else if (lobby.winner == "CATFISH") {
+        end = "Cat Lose";
+      }
+    } else {
+      if (lobby.winner == "CAT") {
+        end = "Catfish Lose";
+      } else if (lobby.winner == "CATFISH") {
+        end = "Catfish Win";
+      }
+    }
```
gchamb: should check if `lobby.winner` is `undefined` before so you can just have if/else instead of elseif

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,217 @@
...
+    </div>
+    <!-- Scenario 2 -->
+  {:else if end == "Cat Lose"}
+    <div class="banner">
+      <h2>You have cat to be kitten me!</h2>
+      {#if catfishes.length > 1}
+        <h2>The impawster, {catfishList} was not caught!</h2>
+      {:else}
+        <h2>The impawsters, {catfishList} were not caught!</h2>
+      {/if}
```
gchamb: if catfishes.length is equal to 1 it will display `The impawsters, {catfishList} were not caught` because 1 isnt greater than 1

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,217 @@
...
+        <h2>{catfishList} were sentenced to 9 lives in purrison!</h2>
+      {:else}
+        <h2>{catfishList} was sentenced to 9 lives in purrison!</h2>
+      {/if}
+      <!-- If the player is the host, show the return button -->
+      {#if isHost}
+        <button
+          on:click={() => {
+            lobbyReturn({ code: lobbyCode });
+          }}>Return to Lobby</button
```
gchamb: you write this multiple times. Can you not render this once?

SBAsai: How would I render it once? An event listener?

nstringham: you can use `on:click|once` like in #107

SBAsai: Oh I think Giantte wanted the whole if statement to only be written once instead of clicked once. It should be fixed.

nstringham: Oh ok

functions/src/winloss.ts
```diff
@@ -0,0 +1,122 @@
...
+  if (lobby.exists === false) {
+    throw new Error("Lobby does not exist!");
+  }
+
+  // loop through each doc in the private player collection via uid and delete them one-by-one
+  const { uids } = lobby.data() as Lobby;
+  for (let i = 0; i <= uids.length; i++) {
+    const playerDoc = privatePlayerCollectionRef.doc(uids[i]);
+    await playerDoc.delete();
+  }
```
nstringham, line 120: an `await` insode a loop can be really slow maybe consider a batch write https://firebase.google.com/docs/firestore/manage-data/transactions#batched-writes

functions/src/winloss.ts
```diff
@@ -0,0 +1,122 @@
...
+export function findWinner(lobbyDocRef: firestore.DocumentReference<Lobby>) {
+  return db.runTransaction(async (transaction) => {
+    const lobby = await transaction.get(lobbyDocRef);
+    const lobbyData = lobby.data();
+
+    if (lobbyData === undefined) {
+      throw new Error("Lobby does not exist!");
+    }
+    const privatePlayerCollection = getPrivatePlayerCollection(lobbyDocRef);
+    const { players, uids } = lobbyData as Lobby;
```
nstringham: don't need this `as`

functions/src/winloss.ts
```diff
@@ -0,0 +1,122 @@
...
+  const lobby = await lobbyDocRef.get();
+  if (lobby.exists === false) {
+    throw new Error("Lobby does not exist!");
+  }
+
+  // loop through each doc in the private player collection via uid and delete them one-by-one
+  const { uids } = lobby.data() as Lobby;
+  for (let i = 0; i <= uids.length; i++) {
+    const playerDoc = privatePlayerCollectionRef.doc(uids[i]);
+    await playerDoc.delete();
```
nstringham: an `await` inside a loop can be really slow maybe consider a batch write https://firebase.google.com/docs/firestore/manage-data/transactions#batched-writes

functions/src/winloss.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    if (lobbyData === undefined) {
+      throw new Error("Lobby does not exist!");
+    }
+    const privatePlayerCollection = getPrivatePlayerCollection(lobbyDocRef);
+    const { players, uids } = lobbyData as Lobby;
+
+    for (let i = 0; i < players.length; i++) {
+      const privatePlayerDocRef = privatePlayerCollection.doc(uids[i]);
+      const privatePlayerDoc = await transaction.get(privatePlayerDocRef);
```
nstringham: an `await` inside a loop can be really slow maybe consider using [Promise.all()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) and [Array.map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map)

functions/src/winloss.ts
```diff
@@ -0,0 +1,122 @@
...
+      const privatePlayerData = privatePlayerDoc.data();
+
+      if (!privatePlayerData) {
+        throw new Error(`Private player not found for uid ${uids[i]}`);
```
nstringham: where are we catching this error?

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,204 @@
...
+  export let lobby: Lobby;
+  export let privatePlayer: PrivatePlayer;
+
+  let end: "Cat Win" | "Cat Lose" | "Catfish Lose" | "Catfish Win" | null = null;
+  let catfishList: string = "";
+  let catfishes: string[];
+  let isHost: boolean;
+  let myRole: Role;
+
+  $: {
```
nstringham: I don't think you want one bug `$:` try to make them as small as posible so they only contain one reactive thing.

for example you could have one that sets the value of `isHost`

nstringham (changes requested):

functions/src/winloss.ts
SBAsai: um I'm throwing the same error as the one in `lobby.ts` so presumably it doesn't need a catch perhaps?

functions/src/winloss.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+  // create a WriteBatch
+  const batch = db.batch();
+  // loop through each doc in the private player collection via uid and delete them one-by-one
+  const { uids } = lobby.data() as Lobby;
+  for (let i = 0; i <= uids.length; i++) {
+    const playerDoc = privatePlayerCollectionRef.doc(uids[i]);
+    // delete each private player document with the WriteBatch
+    batch.delete(playerDoc);
+    batch.commit();
```
nstringham, line 122: the commit should be outside the loop and should be awaited

functions/src/lobby.ts
```diff
@@ -149,12 +150,17 @@ export const onLobbyUpdate = functions.firestore.document("/lobbies/{code}").onU
...
+
   if (lobby.state == "CHAT" && oldLobby.state != "CHAT") {
     const expiration = firestore.Timestamp.fromMillis(
       firestore.Timestamp.now().toMillis() + GAME_STATE_DURATIONS.CHAT * 1000
     );
     lobbyDocRef.set({ expiration }, { merge: true });
   }
+
+  if (lobby.state == "END" && oldLobby.state != "END") {
+    findWinner(lobbyDocRef);
```
nstringham, line 162: this needs to be awaited

functions/src/winloss.ts
nstringham: it probably shouldn't be like that in lobby either but I've bean thinking about this more and I think this is ok for now. we can fix it in sprint 3

functions/src/winloss.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  const { uids, state } = lobby.data() as Lobby;
+  // check if host
+  if (context.auth.uid !== uids[0]) {
+    throw new functions.https.HttpsError("permission-denied", "Not the host of the game!");
+  }
+
+  // check game state
+  if (state !== "END") {
+    throw new functions.https.HttpsError("invalid-argument", "Wrong game state!");
```
nstringham: `invalid-argument` is fine but in the future `failed-precondition` would probobly be better

functions/src/winloss.ts
nstringham: the commit should be outside the loop and should be awaited

functions/src/lobby.ts
nstringham: this needs to be awaited

functions/src/winloss.ts
```diff
@@ -0,0 +1,125 @@
...
+
+// when returning to the lobby, make every player alive again and delete the role, winner, and the private player collection
+export function resetLobby(lobbyDocRef: firestore.DocumentReference<Lobby>) {
+  return db.runTransaction(async (transaction) => {
+    const lobby = await transaction.get(lobbyDocRef);
+    const lobbyData = lobby.data();
+
+    if (lobby === undefined) {
+      throw new Error("Lobby does not exist!");
+    }
```
nstringham: this will never happen because transaction.get never returns undefined

functions/src/winloss.ts
```diff
@@ -0,0 +1,125 @@
...
+    throw new functions.https.HttpsError("permission-denied", "Not the host of the game!");
+  }
+
+  // check game state
+  if (state !== "END") {
+    throw new functions.https.HttpsError("invalid-argument", "Wrong game state!");
+  }
+
+  await resetLobby(lobby.ref);
+  await deletePrivatePlayers(lobby.ref, getPrivatePlayerCollection(lobby.ref));
```
nstringham: each of these functions get the lobby doc wich causes a total of 3 reads. instead lobby should be passed into these functions

client/src/components/WinLoss.svelte
```diff
@@ -0,0 +1,202 @@
...
+  // get the list of catfishes from lobby.ts and put them into a variable
+  catfishes = lobby.players
+    .filter((p) => {
+      return p.role == "CATFISH";
+    })
+    .map((c) => c.displayName);
+  catfishList = formatter.format(catfishes);
+
+  // grab the current user's role
+  myRole = privatePlayer.role;
```
nstringham: this is not reactive but honestly it would probobly be better to get rid of myRole entirly and just use privatePlayer.role on line 39

nstringham (dismissed):

functions/src/winloss.ts
SBAsai: I originally had the commit outside the loop, but the delete wasn't being committed

functions/src/winloss.ts
SBAsai: `getPrivatePlayerCollection` requires a lobby ref document, so would it be fine to leave that as is?

functions/src/winloss.ts
nstringham: > I originally had the commit outside the loop, but the delete wasn't being committed

did you forget to await the commit?

SBAsai: no, I'm testing it right now too and it's not committing the deletes

SBAsai: nvm I had `i <= uids.length` in my for loop and had to remove the `=` lmao

functions/src/winloss.ts
nstringham: > `getPrivatePlayerCollection` requires a lobby ref document, so would it be fine to leave that as is?

I'm not 100% sure what you're asking.

I think you're asking if it is ok to pass lobby and keep `getPrivatePlayerCollection(lobby.ref)` the way it is

and if that's what you are asking then yes

SBAsai: yup, that was all I was wondering. Should be good then.

gchamb (approved):
looks good to me. you addressed nate's changes

nstringham (approved):

Issue #93, opened 2022-10-29
Security for create lobby


nstringham:
currently the create lobby functionality relies on loos security rules and client side database manipulation. make it secure so that catnisses hacker friend dogniss stops making glitched lobbies

tasks:

- [x] move create lobby functionality to server side
- [x] tighten security rules
~~share code with join lobby~~

Acceptance criteria:
a hacker should not be able to make a lobby that a normal user couldn't make

nstringham:
decided it doesn't make sence to share code with join lobby

PR #94, opened 2022-10-29
refactor(functions): ♻️ throw errors instead of returning ErrorResponse

nstringham:
the return type of the callable functions all had issues this PR fixes those issues

this PR also makes functions throw a `HttpsError` instead of returning an object with an `error` property


nstringham:
related to #93 because I want to get this figured out before I share code between create and join lobby.

gchamb (approved):
Looks good to me. Tested it in postman and was getting the appropriate messages

nstringham:
I never used postman for firebase before. What URL do you use?

gchamb:
example

Use the request URL and if you want to use it as an authorized user just take the authorization bearer token and use it in postman

PR #97, opened 2022-11-01
fix: 🔒 lobby security

### LINKED ISSUE #93: Security for create lobby

nstringham:
This PR does 5 things:
1. rename `firestore-functions.ts` → `firebase-functions.ts`
2. make some things in the lobby doc optional that where accidentally made required in #94
3. move create lobby to the server side for better security
4. make a shared constant for the amount of time each game state gets
5. add svelte-check to CI


client/src/routes/+page.svelte
```diff
@@ -44,9 +44,9 @@
       // Create User
       await saveOrCreate(user, userData, name.trim());
       // Create Lobby
-      const code = await createLobby(name.trim());
+      const responce = await createLobby();
```
gchamb: why?

gchamb:
in Chatroom expiration can be undefined now after adding the optional

client/src/routes/+page.svelte
nstringham: why what?

nstringham: the name is no longer an argument because it is read from the userData doc and the return type is a response object that contains a code rather than being the just a string

nstringham: oh wait

nstringham: you mean the spelling

client/src/components/ChatRoom.svelte
```diff
@@ -43,8 +49,12 @@
     }
     // create timer
     timer = setInterval(() => {
-      const diff = Math.floor((lobbyData.expiration.toMillis() - Date.now()) / 1000);
-      countdown = diff;
+      if (lobbyData.expiration == undefined) {
+        countdown = GAME_STATE_DURATIONS.CHAT;
```
gchamb: can you initialize countdown with this value instead of waiting for the interval? We wouldn't have to check if it is undefined on `line 97`.

nstringham: Good idea

gchamb (approved):

zaverbac (commented):
Looks good

mrjoshua520 (approved):

PR #102, opened 2022-11-03
fix(client): 🐛 Display name is blank once user signs out

Muaad2001:
Reload page on sign out function to clear display name once signed out.

### DIFF
--- a/client/src/lib/firebase/auth.ts
+++ b/client/src/lib/firebase/auth.ts
@@ -35,5 +35,6 @@ export async function loginAnonymous() {
 }
 
 export function logOut() {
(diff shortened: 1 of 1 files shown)


client/src/lib/firebase/auth.ts
```diff
@@ -35,5 +35,6 @@ export async function loginAnonymous() {
 }
 
 export function logOut() {
+  location.reload();
```
gchamb, line 38: remove this

gchamb: use browser level naviagtion to reload the page

PR #105, opened 2022-11-03
feat: ✨ Add Leave Lobby Functionality 

### LINKED ISSUE #39: Allow Catniss to leave the lobby
Leave Lobby button for Catniss to click on to take them back to the main page in case she does not want to play in the current lobby.

Catniss can leave the lobby by closing the tab so that she is not forced to press the "leave lobby" button.

This card does not include the case where the browser crashes or runs out of battery

Task List

- [x] Add a leave lobby button to the main lobby page
- [x] Closing the tab should perform the same actions as the button
- [x] Clicking the leave lobby button should remove you from the lobby entirely
- [x] If you are the host then make another player the host 
- [x] If you are the last player in the lobby the lobby should be deleted 
- [x] Create a pop-up that will display if the player tries to leave an in progress game

Acceptance Criteria
- Cantiss can leave a lobby on the lobby page
- Clicking on button takes Catniss back to main page
- Catniss cannot leave while a game is in progress
- If Catniss is the host then another player will be the host so that the lobby does not close when the host leaves.
- if Catniss was the last player in the lobby the lobby should be deleted so that there is not a large sum of empty lobbies constantly running.
- if the game is in progress and the user tries to close the tab it will ask if are user using the [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event)
- if the server side throws an error the client should display it (make this work for start game as well)

mrjoshua520:
closes #39


nstringham:
while playing the game if you reload the page then you leave the lobby. is that acceptable behavior?

client/src/components/Lobby.svelte
```diff
@@ -35,6 +39,23 @@
   function copyLink() {
     navigator.clipboard.writeText(url);
   }
+
+  async function startGame() {
+    try {
+      await findAndStartLobby(lobbyCode);
+    } catch (err) {
+      console.log("HELLLO" + err);
```
nstringham: I doubt you intended to leave that in

client/src/components/Lobby.svelte
```diff
@@ -44,8 +65,21 @@
       <h3>Players: {lobby.players.length}</h3>
     </div>
     <SelectAvatar {lobby} {lobbyCode} />
-    <div class="start">
-      <Button on:click={() => startGame({ code: lobbyCode })}><Label>Start Game</Label></Button>
+    {#if auth.currentUser?.uid === lobby.uids[0]}<div class="start">
+        <Button on:click={() => startGame()}><Label>Start Game</Label></Button>
+      </div>{/if}
```
nstringham: I feel like it would look better if it was like
```svelte
{#if auth.currentUser?.uid === lobby.uids[0]}
  <div class="start">
    <Button on:click={() => startGame()}><Label>Start Game</Label></Button>
  </div>
{/if}
```

client/src/components/Lobby.svelte
```diff
@@ -44,8 +65,21 @@
...
+    <div class="leave">
+      <Button
+        on:click={() => {
+          leaveLobby();
+          goto("/");
```
nstringham, line 75: this makes it go to the home page before it is done leaving the lobby because leaveLobby is not awaited

client/src/components/Lobby.svelte
```diff
@@ -44,8 +65,21 @@
...
+        }}><Label>Leave Lobby</Label></Button
+      >
+    </div>
+    <div class="error">
+      {#if errorMessage !== ""}
+        <p class="error">{errorMessage}</p>
```
nstringham: you probobly don't want 2 things with the same class like that

client/src/components/Lobby.svelte
```diff
@@ -60,7 +94,9 @@
     justify-content: center;
   }
 
-  .start {
+  .start,
+  .leave,
+  .error {
     display: grid;
     place-items: center;
   }
```
nstringham: if these 3 classes are the same then you should only have one class that all 3 elements share

client/src/lib/firebase/start-lobby.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export async function findAndStartLobby(id: string) {
+  // Check if code is valid
+  if (id === "") {
+    throw new Error("Code can't be empty!");
+  }
+
+  // make request to server
+  await startGame({ code: id });
+}
```
nstringham: don't really need this but whatever

client/src/lib/firebase/leave-lobby.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export async function findAndLeaveLobby(id: string) {
+  // Check if code is valid
+  if (id === "") {
+    throw new Error("Code can't be empty!");
+  }
+
+  // make request to server
+  await leaveLobby({ code: id });
+}
```
nstringham: don't really need this but whatever

client/src/routes/game/+page.svelte
```diff
@@ -64,6 +65,13 @@
       // will change privatePlayer to the new doc data
       privatePlayer = doc.data();
     });
+
+    addEventListener("beforeunload", (event) => {
+      event.returnValue = "\\o/";
+      if (lobby?.state === "WAIT" && lobbyCode !== null) {
```
nstringham: this code always runs even if the user clicks "cancel" maybe use the "unload" event

nstringham (changes requested):

client/src/routes/game/+page.svelte
mrjoshua520: We decided not to have the user leave during the unload event so this won't need to be addressed

functions/src/lobby.ts
```diff
@@ -140,6 +140,48 @@ export const joinLobby = functions.https.onCall((data: unknown, context): Promis
...
+
+    // get lobby data
+    const { players, uids } = lobbyInfo.data() as Lobby;
+
+    // Get position of the Player
+    const playerPos = uids.indexOf(auth.uid);
+
+    // If the last player is leaving delete the document instead
+    if (uids.length === 1) {
+      await transaction.delete(lobby);
```
nstringham, line 174: this await doesn't do anything because `transaction.delete` does not return a promise

nstringham: this await doesn't do anything because `transaction.delete` does not return a promise

functions/src/lobby.ts
```diff
@@ -140,6 +140,48 @@ export const joinLobby = functions.https.onCall((data: unknown, context): Promis
...
+    } else {
+      // Remove player from the lobby
+      await transaction.update(lobby, {
```
nstringham: this await doesn't do anything because `transaction.update` does not return a promise

client/src/components/Lobby.svelte
```diff
@@ -44,8 +63,22 @@
       <h3>Players: {lobby.players.length}</h3>
     </div>
     <SelectAvatar {lobby} {lobbyCode} />
-    <div class="start">
-      <Button on:click={() => startGame({ code: lobbyCode })}><Label>Start Game</Label></Button>
+    {#if auth.currentUser?.uid === lobby.uids[0]}<div class="actions">
+        <Button on:click={() => start()}><Label>Start Game</Label></Button>
+      </div>
+    {/if}
```
nstringham: I don't like how the `<div>` is on the same line as `{#if}` but the `</div>` is on a diffrent line from `{/if}`

nstringham (changes requested):

nstringham (dismissed):

nstringham (approved):
approving again after resolving merge conflicts

PR #107, opened 2022-11-03
fix(client): 🐛 Fix the create lobby button making multiple lobbies

### LINKED ISSUE #103: Create lobby button should only happen once
## description
The create lobby button can be clicked more than once before the page loads which creates more than one lobby

## steps to reproduce
1. go to https://cat-me-if-you-can-game.web.app
2. Enter a display name
3. Spam the create lobby button

## expected behavior
only one lobby should be created 

## actual behavior
Multiple lobbies are created

## possible cause
"Once" is not being used to modify the button event 
https://svelte.dev/tutorial/event-modifiers

mrjoshua520:
The create lobby button can no longer be spammed to make multiple lobbies

close #103

### DIFF
--- a/client/src/components/Lobby.svelte
+++ b/client/src/components/Lobby.svelte
@@ -45,7 +45,7 @@
     </div>
     <SelectAvatar {lobby} {lobbyCode} />
     <div class="start">
-      <Button on:click={() => startGame({ code: lobbyCode })}><Label>Start Game</Label></Button>
+      <Button on:click|once={() => startGame({ code: lobbyCode })}><Label>Start Game</Label></Button>
     </div>
     <div class="buttons">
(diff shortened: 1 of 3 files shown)


nstringham:
can we also do this for join lobby and start game?

mrjoshua520:
> can we also do this for join lobby and start game?

Yep, just added it

zaverbac (commented):
Seems like it should work after looking at the documentation for once.

zaverbac (dismissed):
Should work as a solution

gchamb (dismissed):

nstringham:
How about the join button on the join page?

nstringham (approved):

PR #140, opened 2022-11-15
feat: ✨ Delete Lobbies with Cron

### LINKED ISSUE #131: delete old lobbies from firestore
if all the players in a lobby decide to just close the browser at once then the lobby should be deleted eventually instead of taking up space in firestore forever because we don't want to be storing and possibly paying for a ton of old lobbies

Tasks:
- [ ] make expiration non-optional
- [ ] make cron job

Acceptance Criteria:
- lobbies should not be automatically deleted if they are less than a day after the game ends
- lobbies should be automatically deleted within a week of when the game ends
- hackers should not be able to prematurely delete a lobby

gchamb:
How to Test:

- Create a Lobby
- Change the expiration in the lobby doc to 24 hours ago or more
- Click the delete button on the splash screen page (Only there for testing)

### DIFF
--- a/functions/src/cron.ts
+++ b/functions/src/cron.ts
@@ -12,11 +12,11 @@ export const testLogic = functions.https.onRequest(async (req, res) => {
 // export const deleteOldLobbies = functions.pubsub.schedule("15 3 * * *").onRun(async (context) => {
 const deleteOldLobbies = async () => {
   // TODO: Remove if not testing the logic
-  const dayFromTodaySecs = firestore.Timestamp.now().seconds - 86_400; // A day ago from today in seconds
-  const dayFromTodayDate = firestore.Timestamp.fromDate(new Date(dayFromTodaySecs * 1000)); // a day ago from today in date object
+  const oneDayAgoAsSecs = firestore.Timestamp.now().seconds - 24 * 60 * 60; // A day ago from today in seconds
+  const oneDayAgoAsDate = firestore.Timestamp.fromDate(new Date(oneDayAgoAsSecs * 1000)); // a day ago from today in date object
 
   // query all lobby documents that expiration date is either less than or equal from 24 hours from today
-  const oldLobbies = await db.collection("/lobbies").where("expiration", "<=", dayFromTodayDate).get();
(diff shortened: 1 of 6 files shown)


Muaad2001 (dismissed):
checks out

functions/src/lobby.ts
```diff
@@ -53,6 +53,7 @@ export const createLobby = functions.https.onCall(async (data: unknown, context)
     ],
     state: "WAIT",
     alivePlayers: [context.auth.uid],
+    expiration: firestore.Timestamp.now(),
```
nstringham: maybe set this to like 2 or 3 hours instead of instant

nstringham: don't forget to also do this when the game ends and the lobby goes back to WAIT

functions/src/cron.ts
```diff
@@ -0,0 +1,33 @@
...
+export const testLogic = functions.https.onRequest(async (req, res) => {
+  deleteOldLobbies();
+  res.send({ message: "Hi" });
+});
+
+// Schedule for every day at 3:15AM
+// export const deleteOldLobbies = functions.pubsub.schedule("15 3 * * *").onRun(async (context) => {
+const deleteOldLobbies = async () => {
+  // TODO: Remove if not testing the logic
+  const dayFromTodaySecs = firestore.Timestamp.now().seconds - 86_400; // A day ago from today in seconds
```
nstringham: I would prefer if this was `24 * 60 * 60`

functions/src/cron.ts
```diff
@@ -0,0 +1,33 @@
...
+  const dayFromTodayDate = firestore.Timestamp.fromDate(new Date(dayFromTodaySecs * 1000)); // a day ago from today in date object
```
nstringham: let's call these constants something that if obvious to everone that it refers to the past. maybe we could call it `oneDayAgo` or `twentyFourHoursAgo`

nstringham (dismissed):
looks pretty good don't forget to remove the testing things

mrjoshua520 (approved):

nstringham (approved):

PR #154, opened 2022-11-22
fix(client):🐛Fix lobby chat issue that doesn't let users leave lobby

### LINKED ISSUE #153: Fix leave lobby bug
Users aren't allowed to leave game. This is because of LobbyChat.svelte. The user does leave the lobby on the backend.

to recreate the error:
1. create a lobby
2. join the lobby with a new account and the new user shouldn't be able to leave the lobby.

Muaad2001:
Users can now leave the lobby.

### DIFF
--- a/client/src/routes/game/+page.svelte
+++ b/client/src/routes/game/+page.svelte
@@ -99,7 +99,7 @@
     <div class="spinner-wraper">
       <CircularProgress indeterminate />
     </div>
-  {:else if !userInfo.alive}
+  {:else if !lobby.alivePlayers.includes($user.uid)}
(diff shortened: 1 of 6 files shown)


client/src/routes/game/+page.svelte
```diff
@@ -119,6 +124,7 @@
   {:else}
     unknown lobby state: {lobby.state}
   {/if}
+  <!-- put lobbychat were alive users will be able to access it -->
```
njumaah: what is this

njumaah (changes requested):

njumaah (approved):

PR #157, opened 2022-11-22
feat(client): ✨ Show users answer on prompt page when they click done

### LINKED ISSUE #156: Show confirmation that Catniss has submitted their answer.

Muaad2001:
When user hits done on prompt page their answer will be displayed, If they start typing another answer it won't update the answer until they hit done.


client/src/components/Prompt.svelte
```diff
@@ -76,7 +82,10 @@
     <Textfield input$id="prompt-answer" bind:value={answer} bind:dirty invalid={dirty && error != undefined} required>
       <HelperText validationMsg slot="helper">{error ?? ""}</HelperText>
     </Textfield>
-    <Button type="submit" disabled={error != undefined}><Label>Done</Label></Button>
+    <Button type="submit" disabled={error != undefined} on:click={showAnswer}><Label>Done</Label></Button>
```
nstringham: this should be in the submit event listner on the form

client/src/components/Prompt.svelte
```diff
@@ -76,7 +82,10 @@
...
+    {#if showUserAnswer}
```
nstringham: why not just do `displayAnswer != ""` or even make displayAnswer start as undefined and display if it is not undefined

nstringham (changes requested):
I think this would be better if it was doing database reads like we do for voting but I guess it's ok like this

zaverbac (approved):

SBAsai:
I think that when the timer hits 0:00, you shouldn't be able to keep entering in answers and have `Your Answer:` pop up when the answer isn't saved

nstringham:
> I think that when the timer hits 0:00, you shouldn't be able to keep entering in answers and have `Your Answer:` pop up when the answer isn't saved

I like this idea but it should probably be another card

I created #166 which would resolve this

nstringham (approved):

nstringham:
Requesting @gchamb's review because I think he said he wanted this to be a database subscription

nstringham:
merging this and moving the firebase reads to #167

PR #160, opened 2022-11-29
feat(functions): ✨ Trim ending spaces off name, no error throw.

### LINKED ISSUE #129: Automatically Trim DisplayName on join and splash
Description: When Catniss enters here display name with a space at the end and hits enter, it is automatically trimmed because we do not want there to be an error thrown that there is a a trailing whitespace.

Tasks:

- [ ] change validation on join
- [ ] change validation on splash
- [ ] change backend call on join
- [ ] change backend call on splash

Acceptance criteria:
- entering a name ending in a space will not show an error
- clicking the join button will store a trimmed version into firestore

cmcclusky:
Make it so that you may enter a name with spaces at the end. The code will trim it for you and won't throw an error. This works for both splash and join and does still trim all the ending spaces off.

### DIFF
--- a/functions/src/firestore-types/users.ts
+++ b/functions/src/firestore-types/users.ts
@@ -36,11 +36,6 @@ export type UserData = {
 };
 
 export function displayNameValidator(displayName: string): { valid: true } | { valid: false; reason: string } {
-  let newname = displayName;
-  while (newname.charAt(newname.length - 1) === " ") {
-    newname = newname.slice(0, -1);
-  }
-  displayName = newname;
   if (displayName === "") {
     return { valid: false, reason: "Display name must not be empty" };
   }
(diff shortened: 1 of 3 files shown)


gchamb (dismissed):
fix the CI/CD check

functions/src/firestore-types/users.ts
```diff
@@ -36,6 +36,11 @@ export type UserData = {
 };
 
 export function displayNameValidator(displayName: string): { valid: true } | { valid: false; reason: string } {
+  let newname = displayName;
+  while (newname.charAt(newname.length - 1) === " ") {
+    newname = newname.slice(0, -1);
+  }
+  displayName = newname;
   if (displayName === "") {
```
gchamb: you want to trim the name in the parameter not in this function. Change this back to the original

functions/src/firestore-types/users.ts
```diff
@@ -48,7 +53,7 @@ export function displayNameValidator(displayName: string): { valid: true } | { v
   }
 
   if (displayName !== displayName.trim()) {
-    return { valid: false, reason: "Display name must not contain leading or trailing whitespace" };
+    return { valid: false, reason: "Display name must not contain leading whitespace" };
```
gchamb: change back

gchamb (approved):

nstringham:
~~In addition to trimming before passing to the validator you should trim before passing to the backend~~

Edit: nevermind we're already doing that

Muaad2001 (approved):

PR #162, opened 2022-11-29
feat(client): 💄 Center error text in lobby.svelte

njumaah:


zaverbac (commented):
I think it looks much better centered. I wonder if the padding should be increased so it's spaced out a little further from the avatars

njumaah:
> I think it looks much better centered. I wonder if the padding should be increased so it's spaced out a little further from the avatars

played around with the padding; 20px seems the best for desktop and mobile

njumaah:
image

zaverbac (approved):
I think it looks good

mrjoshua520 (approved):

PR #170, opened 2022-12-03
feat(client): ✨ Add Skip as voting option

### LINKED ISSUE #134: Skip vote and reduce timer when everyone votes
Catniss should be able to click and "Abstain" button to skip the vote because maybe they want to spend more time gathering information during the game

The timer for voting should be reduced once everyone has entered their vote because having to wait the whole time would get annoying.

Task List
- [x] the type of uid is `string | null` on the vote in firestore for abstaining vote.
- [x] Make "Skip Vote" button in frontend.
- [x] Check if skip votes is >= any other option in backend.
- [x] Make timer shorter when all players have voted in backend.

Acceptance Criteria
- Timer goes to set value when everyone has either voted or abstained
- the abstain button works
- if more people obtained than voted for a particular player then no on is voted out
- hacker should not be able go to the reduced the timer before everyone has voted

njumaah:
close: #134


Muaad2001 (dismissed):
looks good

mrjoshua520 (commented):
Everything looks good to me so this is kinda an approval but I personally think the skip vote button being centered below the players grid would look better. I think we should ask the group before merging

mrjoshua520 (approved):

functions/src/lobby.ts
```diff
@@ -439,20 +440,41 @@ export const onVoteWrite = functions.firestore.document("/lobbies/{code}/votes/{
       return;
     }
 
-    const { players, uids } = lobbyData;
+    const { players, uids, skipVote: oldSkipVote } = lobbyData;
+    let skipVote = oldSkipVote;
```
gchamb, line 444: why create an alias if you create another variable with the same name?

gchamb (dismissed):
left a minor comment above. then ill merge it

functions/src/lobby.ts
nstringham: I think this is done so that it is mutable but maybe it would be better to do a separate deconstruction with let

njumaah: i think its fine the way it is

gchamb (approved):
left a minor comment above. then ill merge it

PR #174, opened 2022-12-05
fix: await updating alive players

nstringham:
this is needed to stop firebase from killing the server prematurely


Muaad2001 (approved):
looks good

mrjoshua520 (approved):

PR #175, opened 2022-12-05
feat(client): CSS changes to prompt and game

### LINKED ISSUE #136: Add margins on prompts
Change CSS so that prompts don't get so close to the side of the screen (on mobile). This issue particularly shows when the prompt is long.

This is because it will look nicer for the user and be a better user experience overall.

Tasks:
- [x] make look pretty :)

Acceptance Criteria
- Prompts don't get too close to side of screen on mobile view.
- Make sure it also still looks good on desktop view.

nstringham:
This PR fixes #136 but also changes other CSS stuff that is nearby/related

Oops I accidentally committed the new images to this branch

image

image

image


mrjoshua520 (dismissed):
These might be easier to explain in a VC but while running this PR I noticed these things.

1. The receive role screen didn't appear on one of my browsers (Brave not private) however the others (brave private, edge not private and firefox not private all showed up fine
2. Same thing happened with the win screen with the same browsers acting the same ways
3. The chatting screen on edge, private brave and firefox all looked great however on non-private brave the following happened: The header was part of the page that could be scrolled, the chat message entry box was pushed off the page once it ran out of room, and when the page loaded it started in the center.

Attached are some images of the chat page. Let me know if you want to see a demo of this happening and it very well could be on my browsers end


This is the way the page started out
Starting look

This is the page after a few messages were posted
ending look

mrjoshua520 (approved):

PR #176, opened 2022-12-05
refactor: 🎨 Refactor the players array into a map

### LINKED ISSUE #130: refactor players into a map
the players array is hard to use because indexof is O(n) we should replace it with a map from uid to player because this is faster more readable and works in firestore function

Task List:
- [x] Update the players in firestore
- [x] Update all locations where players is being used
- [x] All Client Files
- [x] All Functions Files
- [x] All Firestore Rules

Acceptance criteria:
- players should be `{[uid: string]: Player}` instead of `Player[]`
- everything else should still work

mrjoshua520:
Keep in mind this refactor also removes the UID array


nstringham:
Are we still planning to remove alivePlayers?

client/src/components/Lobby.svelte
```diff
@@ -81,16 +82,16 @@
...
     <h3>Code: {lobbyCode}</h3>
     <h3>Players: {lobby.players.length} / 8</h3>
-    {#if $user?.uid === lobby.uids[0]}
+    {#if $user?.uid === lobby.host}
       <div class="settings">
         <LobbySettings {lobby} {lobbyCode} />
       </div>
     {/if}
-    {#if lobby.players.length < minPlayers}
+    {#if Object.keys(lobby.players).length < minPlayers}
```
nstringham: why isn't this `playersLength`?

client/src/components/Lobby.svelte
```diff
@@ -99,11 +100,9 @@
     {/if}
   </div>
   <SelectAvatar {lobby} {lobbyCode} on:change={(event) => onAvatarSelect(event.detail.value)} />
-  {#if $user?.uid === lobby.uids[0]}
+  {#if $user?.uid === lobby.host}
     <div class="actions">
-      <Button on:click={start} disabled={lobby.players.length < minPlayers || waiting}>
-        <Label>Start Game</Label>
-      </Button>
+      <Button on:click|once={() => start()} disabled={playersLength < minPlayers}><Label>Start Game</Label></Button>
```
nstringham: are you sure you want to add this `once` and remove the `waiting`?

client/src/components/SelectAvatar.svelte
```diff
@@ -31,18 +31,14 @@
     }
 
     if (lobby != undefined) {
-      for (const player of lobby.players) {
-        newAvatarChoices[player.avatar - 1].displayName = player.displayName;
-        newAvatarChoices[player.avatar - 1].available = false;
-        const playerIndex = lobby.players.indexOf(player);
-        newAvatarChoices[player.avatar - 1].uid = lobby.uids[playerIndex];
+      for (const uid in lobby.players) {
```
nstringham, line 33: you should probobly either do Object.entries or do `const player = lobby.players[uid]` so the we don't have to say `lobby.players[uid]` every time because it's kinda long

client/src/components/Vote.svelte
```diff
@@ -33,24 +33,24 @@
 <div class="voting">
   <p class="mdc-typography--headline4">Vote out the catfish</p>
   <div class="voting-grid">
-    {#each lobby.players as { avatar, displayName, votes, alive, promptAnswer }, i}
-      <div class="vote-container {!alive ? 'dead' : ''}">
+    {#each Object.entries(lobby.players) as [uid, player]}
```
nstringham: you should replace `player` with `{ avatar, displayName, votes, alive, promptAnswer }` so that the sorter variable names can be used like before this PR

client/src/routes/game/+page.svelte
```diff
@@ -58,9 +58,8 @@
       errorToJoin("Lobby doesnt exist!");
       return;
     }
-
-    // check for user being null so privatePlayerDocRef can work
-    if ($user === null || !lobby.uids.includes($user.uid)) {
+    // if the user isn't signed in or not apart of this lobby then redirect them
+    if ($user === null || !Object.keys(lobby.players).includes($user.uid)) {
```
nstringham: this could use `in` instead of `Object.keys().includes()`

client/src/routes/game/+page.svelte
```diff
@@ -111,7 +110,7 @@
   }
 
   // Reactive Calls
-  $: if (lobby !== undefined && $user !== null && !lobby.uids.includes($user.uid)) {
+  $: if (lobby !== undefined && $user !== null && !Object.keys(lobby.players).includes($user.uid)) {
```
nstringham: this could use `in` instead of `Object.keys().includes()`

firestore.rules
```diff
@@ -10,11 +10,11 @@ service cloud.firestore {
       }
 
       function inLobby(){
-        return request.auth.uid in lobbyData().uids;
+        return request.auth.uid in lobbyData().players;
       }
 
       function isAlive(){
-        return request.auth.uid in lobbyData().alivePlayers;
+        return request.auth.uid in lobbyData().alivePlayers;
```
nstringham: I don't think you want to add this space

functions/src/lobby.ts
```diff
@@ -156,8 +158,8 @@ export const joinLobby = functions.https.onCall((data: unknown, context): Promis
     const userInfo = user.data() as UserData;
 
     // get lobby data
-    const { players, uids, bannedPlayers } = lobbyInfo.data() as Lobby;
-    if (uids.includes(auth.uid)) {
+    const { players, bannedPlayers } = lobbyInfo.data() as Lobby;
+    if (Object.keys(players).includes(auth.uid)) {
```
nstringham: this could use `in` instead of `Object.keys().includes()`

functions/src/lobby.ts
```diff
@@ -193,14 +195,17 @@ export const joinLobby = functions.https.onCall((data: unknown, context): Promis
       });
     } else {
       // add player
+      const existingPlayers = players;
```
nstringham: this is redundant just use players

functions/src/lobby.ts
```diff
@@ -230,19 +235,41 @@ export const leaveLobby = functions.https.onCall((data: unknown, context): Promi
...
+    const { players, host } = lobbyInfo.data() as Lobby;
 
-    // Get position of the Player
-    const playerPos = uids.indexOf(auth.uid);
+    const newPlayers: Lobby["players"] = {};
+    for (const uid in players) {
+      if (uid !== auth.uid) {
+        newPlayers[uid] = players[uid];
+      }
+    }
```
nstringham: why not just use `delete`?

functions/src/lobby.ts
```diff
@@ -230,19 +235,41 @@ export const leaveLobby = functions.https.onCall((data: unknown, context): Promi
...
+        newHost = auth.uid;
+      }
+
       // Remove player from the lobby
       transaction.update(lobby, {
-        players: firestore.FieldValue.arrayRemove(players[playerPos]),
-        uids: firestore.FieldValue.arrayRemove(auth.uid),
+        players: newPlayers,
+        host: newHost ?? host,
       });
```
nstringham: is this the same as in `kick-ban.ts` ? if it is then it should probobly be a helper function to reduce code duplication

functions/src/result.ts
```diff
@@ -11,17 +12,19 @@ export async function determineWinner(lobbyDoc: DocumentSnapshot<Lobby>, transac
     throw new Error("Lobby not found");
   }
 
-  const { uids } = lobbyData;
   const { players: currentPlayers } = lobbyData;
   let winner: Role | undefined;
-  const alteredPlayers: Player[] = JSON.parse(JSON.stringify(currentPlayers));
+  const alteredPlayers = { ...currentPlayers };
```
nstringham: this is not a deep copy

functions/src/winloss.ts
```diff
@@ -46,19 +46,19 @@ export async function endGameProcess(
   transaction: firestore.Transaction
 ) {
   // apply the stats
-  let { players } = lobbyData;
-  const { uids, winner } = lobbyData;
+  const { players } = lobbyData;
+  const { winner } = lobbyData;
```
nstringham: these lines can be combined

functions/src/winloss.ts
```diff
@@ -46,19 +46,19 @@ export async function endGameProcess(
...
 
   // update each players doc
   await Promise.all(
-    uids.map(async (uid, index) => {
+    Object.keys(lobbyData.players).map(async (uid) => {
```
nstringham: consider Object.entries instead of keys so you don't have to do lobbyData.players[uid] a bunch

functions/src/vote.ts
```diff
@@ -3,30 +3,40 @@ import type { DocumentReference, Transaction } from "firebase-admin/firestore";
...
 export function findVoteOff(lobbyData: Lobby, lobbyDocRef: DocumentReference<Lobby>, transaction: Transaction) {
-  const { players, uids } = lobbyData;
+  const { players } = lobbyData;
   let votedOff: string | undefined = undefined;
 
   // find the player with the most votes
-  const [most, secondMost]: Player[] = JSON.parse(JSON.stringify(players)).sort((a: Player, b: Player) => {
-    return b.votes - a.votes;
-  });
+  let most: Player | null = null;
```
nstringham: should probobly make this a uid that way you don't need to do that stuff with `mostUID`

nstringham (changes requested):

client/src/components/Lobby.svelte
```diff
@@ -96,11 +97,11 @@
     {/if}
   </div>
   <SelectAvatar {lobby} {lobbyCode} on:change={(event) => onAvatarSelect(event.detail.value)} />
-  {#if $user?.uid === lobby.uids[0]}
+  {#if $user?.uid === lobby.host}
     <div class="actions">
-      <Button on:click={start} disabled={lobby.players.length < minPlayers || waiting}>
-        <Label>Start Game</Label>
-      </Button>
+      <Button on:click={() => start()} disabled={playersLength < minPlayers || waiting}
```
nstringham, line 104: why was this changed from `start` to `() => start()`?

firestore.rules
```diff
@@ -87,7 +87,7 @@ service cloud.firestore {
         function checkTarget(){
           return request.resource.data.keys().hasOnly(['target'])
             && (request.resource.data.target == null || (request.resource.data.target is string
-            && request.resource.data.target in lobbyData().alivePlayers));
+            && lobbyData().players[uid].alive));
```
nstringham: this is checking if the person casting the vote is alive but it should be checking if the person they are voting for is alive

functions/src/chat.ts
```diff
@@ -59,8 +59,7 @@ export async function setAndDeleteAnswers(
     const promptData = promptAnswerDoc.data();
     const promptAns = promptData.answer;
 
-    const playerIndex = uids.indexOf(promptAnswerDoc.id);
-    players[playerIndex].promptAnswer = promptAns;
+    players[promptAnswerDoc.id].promptAnswer = promptAns;
```
nstringham: the `promptAns` constant is only used once you should consider inlining it

functions/src/firestore-types/lobby.ts
```diff
@@ -137,20 +130,24 @@ export type Lobby = {
...
-  lobbySettings: LobbySettings;
+  host: string;
 
-  /*
+  /**
    * array of uids of banned players
    */
   bannedPlayers: string[];
+
+  /*
```
nstringham: can you add a second `*` here so that it becomes a JSDoc

functions/src/firestore-types/lobby.ts
```diff
@@ -137,20 +130,24 @@ export type Lobby = {
   votedOff?: string | "NONE";
 
   /**
-   *
    * number of people who skipped voting
    */
   skipVote: number;
 
   /**
    * settings that can be edited in the lobby
```
nstringham, line 138: this comment is doesn't seam righ

functions/src/lobby.ts
```diff
@@ -105,17 +105,18 @@ export const startGame = functions.https.onCall(async (data: unknown, context):
...
     }
+
     // check if the request is coming from the host of the game
-    const { uids, lobbySettings } = lobby.data() as Lobby;
+    const { players, lobbySettings } = lobby.data() as Lobby;
 
     const minPlayers = lobbySettings.catfishAmount * 2 + 2;
 
-    if (auth.uid !== uids[0]) {
+    if (auth.uid !== lobby.data()?.host) {
```
nstringham: I don't like the duplicate calls to `data()` consider adding `host` to line 110

functions/src/lobby.ts
```diff
@@ -241,19 +235,32 @@ export const leaveLobby = functions.https.onCall((data: unknown, context): Promi
...
+      if (auth.uid == host) {
+        for (const uid in players) {
+          const currentPlayerTimeJoined = players[uid].timeJoined;
+          if (currentPlayerTimeJoined.toMillis() < earliestJoinedTime.toMillis()) {
+            earliestJoinedTime = currentPlayerTimeJoined;
+            newHost = uid;
+          }
+        }
+      }
+
```
nstringham: this code is duplicated in kick and ban I recomend pulling it out into a function to reduce code duplication.

functions/src/kick-ban.ts
```diff
@@ -21,15 +22,43 @@ export const kick = functions.https.onCall((data: unknown, context): Promise<voi
...
-      uids: firestore.FieldValue.arrayRemove(data.uid),
-    });
+    const { players, host } = lobbyData as Lobby;
+    const newPlayers: Lobby["players"] = {};
+
+    for (const uid in players) {
+      if (uid !== data.uid) {
+        newPlayers[uid] = players[uid];
+      }
+    }
```
nstringham: we could probobly get rid of `newPlayers` like we did in `leaveLobby`

functions/src/result.ts
```diff
@@ -11,17 +11,15 @@ export async function determineWinner(lobbyDoc: DocumentSnapshot<Lobby>, transac
     throw new Error("Lobby not found");
   }
 
-  const { uids } = lobbyData;
-  const { players: currentPlayers } = lobbyData;
+  const { players } = lobbyData;
   let winner: Role | undefined;
-  const alteredPlayers: Player[] = JSON.parse(JSON.stringify(currentPlayers));
```
nstringham, line 17: does getting rid of altered players make the role show up in the lobby when the game is not over yet?

functions/src/vote.ts
```diff
@@ -1,32 +1,34 @@
...
+  let most: string | null = null;
+  let mostValue = 0;
+
+  for (const uid in lobbyData.players) {
+    if (players[uid].votes > mostValue || players[uid].votes > lobbyData.skipVote) {
+      most = uid;
+      mostValue = players[uid].votes;
+    } else if (players[uid].votes == mostValue || players[uid].votes == lobbyData.skipVote) {
+      most = null;
     }
```
nstringham: I think you don't want to have the || skipVote here because it means that any player that has more votes than skip is voted off.

Why not just initlaize `let mostValue = lobbyData.skipVote` ?

nstringham: I think you don't want to have the || skipVote here because it means that any player that has more votes than skip is voted off.

Why not just `let mostValue = lobbyData.skipVote` ?

functions/src/vote.ts
```diff
@@ -1,32 +1,34 @@
...
+  }
+
+  // if they're not equal then most is voted off
+  if (most != null) {
     // replace the shallow one copy object with the deep copy object changes
```
nstringham: is this comment still correct?

functions/src/vote.ts
```diff
@@ -1,32 +1,34 @@
...
-    if (mostIndex != undefined) {
-      players[mostIndex].alive = false;
-      // save their name
-      votedOff = uids[mostIndex];
+    if (most != undefined) {
```
nstringham: why are we checking if most is undefined?

functions/src/winloss.ts
```diff
@@ -28,7 +28,7 @@ export const lobbyReturn = functions.https.onCall(async (data: unknown, context)
 
     const lobbyData = lobbyDoc.data() as Lobby;
     // check if host
-    if (auth.uid !== lobbyData.uids[0]) {
+    if (auth.uid === lobbyData.host) {
```
nstringham: are you sure you want to change it from `!==` to `===`?

functions/src/winloss.ts
```diff
@@ -49,19 +49,17 @@ export async function endGameProcess(
...
+    Object.entries(lobbyData.players).map(async ([uid, player]) => {
       const userDocRef = userCollection.doc(uid);
       const userDoc = await transaction.get(userDocRef);
 
-      if (players[index].role == undefined) {
-        functions.logger.error("This player's role doesn't exist!");
+      if (player.role == undefined) {
         return;
-      } else if (players[index].role == "SPECTATOR") {
+      } else if (players[uid].role == "SPECTATOR") {
```
nstringham: why not just use `player`?

functions/src/winloss.ts
```diff
@@ -49,19 +49,17 @@ export async function endGameProcess(
...
 
   // update each players doc
   await Promise.all(
-    uids.map(async (uid, index) => {
+    Object.entries(lobbyData.players).map(async ([uid, player]) => {
       const userDocRef = userCollection.doc(uid);
       const userDoc = await transaction.get(userDocRef);
 
-      if (players[index].role == undefined) {
-        functions.logger.error("This player's role doesn't exist!");
```
nstringham, line 62: why is the log message removed?

nstringham (dismissed):

client/src/components/Lobby.svelte
mrjoshua520: That's only changing due to a merge from dev. It's been like that on since shunsuke did the once button fix. That's currently how it is on dev/prod as well

functions/src/winloss.ts
mrjoshua520: I might not be understanding what you mean but putting "player" in place of "uid" gives me "Type 'Player' cannot be used as an index type". Above I am using object.entries and getting the uid and player map as the key and value. I just assumed I had to use uid since that is the key we want

nstringham: I mean `player` instead of `players[uid]`

mrjoshua520: ohhh gotcha okay

nstringham:
@mrjoshua520 some of my comments got resolved but I don't see any new commits am I missing something?

mrjoshua520:
> @mrjoshua520 some of my comments got resolved but I don't see any new commits am I missing something?

There are commits waiting to be pushed but I figured I wouldn't until I got a chance to fix some of the bugs that are still in the branch

functions/src/result.ts
mrjoshua520: It does yes. It's been added back to avoid that

functions/src/util.ts
```diff
@@ -11,11 +12,12 @@ export function generatePairs(lobbyData: Lobby): GeneratedPairs {
...
-  lobbyData.players.forEach((player, playerIndex) => {
-    if (player.alive) {
-      aliveUids.push(lobbyData.uids[playerIndex]);
+
+  for (const uid in lobbyData.players) {
+    if (lobbyData.players[uid].alive) {
+      aliveUids.push(uid);
     }
-  });
+  }
```
nstringham, line 20: this could be done functionally with `filter` but this is ok

functions/src/util.ts
```diff
@@ -33,3 +35,19 @@ export function generatePairs(lobbyData: Lobby): GeneratedPairs {
 
   return { pairs, stalker };
 }
+
+export function updateHost(players: { [uid: string]: Player }) {
+  let newHost: string | undefined;
+  let earliestJoinedTime = Timestamp.now();
```
nstringham: it would be slightly faster to store the millis derectly in earliestJoinedTime but if you think this is more readable then it's worth it

functions/src/lobby.ts
```diff
@@ -234,19 +229,19 @@ export const leaveLobby = functions.https.onCall((data: unknown, context): Promi
...
     } else {
       // Remove player from the lobby
       transaction.update(lobby, {
-        players: FieldValue.arrayRemove(players[playerPos]),
-        uids: FieldValue.arrayRemove(auth.uid),
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Workaround for https://github.com/googleapis/nodejs-firestore/issues/1808
+        players: players satisfies Lobby["players"] as any,
+        host: updateHost(players) ?? host,
       });
     }
```
nstringham: this still looks like it has some duplicaed code with kick and ban

functions/src/lobby.ts
```diff
@@ -234,19 +229,19 @@ export const leaveLobby = functions.https.onCall((data: unknown, context): Promi
...
+    if (Object.keys(players).length === 0) {
       transaction.delete(lobby);
     } else {
       // Remove player from the lobby
       transaction.update(lobby, {
-        players: FieldValue.arrayRemove(players[playerPos]),
-        uids: FieldValue.arrayRemove(auth.uid),
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Workaround for https://github.com/googleapis/nodejs-firestore/issues/1808
+        players: players satisfies Lobby["players"] as any,
+        host: updateHost(players) ?? host,
```
nstringham: I don't think you want `?? host` because it could be the host that is leaving

functions/src/result.ts
```diff
@@ -11,17 +11,16 @@ export async function determineWinner(lobbyDoc: DocumentSnapshot<Lobby>, transac
     throw new Error("Lobby not found");
   }
 
-  const { uids } = lobbyData;
-  const { players: currentPlayers } = lobbyData;
+  const { players } = lobbyData;
   let winner: Lobby["winner"];
-  const alteredPlayers: Player[] = JSON.parse(JSON.stringify(currentPlayers));
+  const alteredPlayers: Player[] = JSON.parse(JSON.stringify(players));
```
nstringham, line 16: I don't think you can do this anymore because Player now contains a timestamp

nstringham (changes requested):

PR #186, opened 2022-12-07
fix(client): 🐛 respond to updates to vote user off

nstringham:
this fixes a bug where the inital render would try to get `undifined.webp`


njumaah (approved):
looks like if votedOffUser is undefined, it no longer tries to get an undefined.webp and instead just displays no one was voted off

mrjoshua520 (approved):

PR #187, opened 2022-12-07
fix(client): 🥅 handle errors for verifyExpiration

nstringham:
the main benefit of this PR is that `Too early to make request` is no longer an error printed to the console so it is easier to diagnose problems


njumaah (commented):
how would i be able to test this?

nstringham:
just run the app and look in the browser console. it should show info instead of error (grey not red)

mrjoshua520 (approved):

njumaah (approved):

PR #189, opened 2022-12-08
fix(functions): 🐛 stalker prompt shows and no multiple stalkers

### LINKED ISSUE #161: Stalker's prompt answer gets deleted/does not save
For some reason, the stalker's prompt answer never shows during the voting phase. (my assumption would be that we are getting the prompt answers from the chat phase instead of the prompt phase)

In chat.ts it looks like prompt is being stored once the chat rooms are deleted.

gchamb:
This shows the stalker's prompt answer and fixes the multiple stalker/spectator problem.

ignore the branch name should just be gchamb/fix-stalker

### DIFF
--- a/functions/src/chat.ts
+++ b/functions/src/chat.ts
@@ -28,7 +28,9 @@ export async function deleteChatCollections(lobbyDoc: firestore.DocumentReferenc
   for (const chatRoomDoc of chatRooms.docs) {
     // place messages in batch
     const chatMessages = await getChatRoomMessagesCollection(chatRoomDoc.ref).get();
-    chatMessages.docs.map((messageDoc) => batch.delete(messageDoc.ref));
+    for (const messageDoc of chatMessages.docs) {
+      batch.delete(messageDoc.ref);
+    }
 
     // place room in batch
     batch.delete(chatRoomDoc.ref);
@@ -37,7 +39,7 @@ export async function deleteChatCollections(lobbyDoc: firestore.DocumentReferenc
   await batch.commit();
 }
 
-export async function setAnswers(lobbyData: Lobby, lobbyDoc: DocumentReference<Lobby>, transaction: Transaction) {
+export async function setAndDeleteAnswers(lobbyData: Lobby, lobbyDoc: DocumentReference<Lobby>, transaction: Transaction) {
   const { players, uids } = lobbyData;
   const promptAnswersSnaphot = await transaction.get(getPromptAnswerCollection(lobbyDoc));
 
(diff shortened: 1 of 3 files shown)


mrjoshua520 (dismissed):
Looks great! Someone else should look at it before merging though

nstringham (dismissed):
we decided we should not delete prompt answers at the end pf prompt and use that at the beginning of vote and we should also delete chat messages asyncrounstly after setting the state t vote

functions/src/lobby.ts
```diff
@@ -510,7 +523,13 @@ export const verifyExpiration = functions.https.onCall(async (data, context): Pr
       await determineWinner(lobbyDoc, transaction);
     }
   });
+
+  if (lobby?.state == "CHAT") {
+    await deleteChatCollections(lobbyDocRef);
+  }
+
   if (lobby?.state == "VOTE") {
+    await deletePromptAnswers(lobbyDocRef);
```
nstringham: why are prompt answers being deleted twice?

functions/src/chat.ts
```diff
@@ -19,44 +20,50 @@ export async function deleteLobbyChatMessages(lobbyDoc: DocumentReference<Lobby>
...
-export async function deleteChatRooms(lobbyData: Lobby, lobbyDoc: DocumentReference<Lobby>, transaction: Transaction) {
+export async function deleteChatCollections(lobbyDoc: firestore.DocumentReference<Lobby>) {
+  const batch = db.batch();
+
+  const chatRooms = await getChatRoomCollection(lobbyDoc).get();
+
+  for (const chatRoomDoc of chatRooms.docs) {
+    // place messages in batch
+    const chatMessages = await getChatRoomMessagesCollection(chatRoomDoc.ref).get();
+    chatMessages.docs.map((messageDoc) => batch.delete(messageDoc.ref));
```
nstringham: this does not return anything so it should be a map.

maybe conside a for of loop

functions/src/chat.ts
```diff
@@ -19,44 +20,50 @@ export async function deleteLobbyChatMessages(lobbyDoc: DocumentReference<Lobby>
...
-    transaction.delete(room);
-  }
+  // iterate through promptAnswer docs and place the answer on their player object
+  for (const promptAnswerDoc of promptAnswersSnaphot.docs) {
+    const promptData = promptAnswerDoc.data();
+    let promptAns = promptData.answer;
 
-  for (const stalker of stalkers.docs) {
-    transaction.update(stalker.ref, { stalker: false });
+    if (promptAns === "" || !promptAnswerDoc.exists) {
```
nstringham: don't need to check `!promptAnswerDoc.exists` because querries already garentee that the documents exist

functions/src/chat.ts
```diff
@@ -19,44 +20,50 @@ export async function deleteLobbyChatMessages(lobbyDoc: DocumentReference<Lobby>
...
+      promptAns = "no answer";
+    }
+
+    const playerIndex = uids.indexOf(promptAnswerDoc.id);
+    players[playerIndex].promptAnswer = promptAns;
```
nstringham, line 63: I feel like this could just be `players[playerIndex].promptAnswer = promptData.answer ?? "no answer"`

nstringham (dismissed):
one disavantage of what we are doing here iswe read the prompt awnsers twice at the end of chat once in `setAnswers` and once in `deletePromptAnswers` we could store if and reuse it but I think this is probobly fine for now

nstringham (approved):

mrjoshua520 (approved):

PR #244, opened 2022-12-22
perf(client): ⚡ Cache static assets

nstringham:
This PR changes URLs like `/images/winloss/catfish-win.webp` to URLs like `/_app/immutable/assets/catfish-win-a02ca162.webp` the `a02ca162` at the end is a file hash so if we change the file contents we will also change it's URL. this means that we can tell the browser that the file at that URL will never change so it doesn't need to request it ever time it visits the page it can just use the one it cached last time the user visited the page.


gchamb (approved):
Looks good. Idk if you wanna do it in this PR but there's two different height values for the body tag element

nstringham:
> there's two different height values for the body tag element

this was intentional because not all browsers supported the `dvh` unit so `vh` is a fallback

that being said Chrome has now added support for `dvh` so maybe we should consider removing the fallback
https://caniuse.com/mdn-css_types_length_viewport_percentage_units_dynamic

gchamb:
> > there's two different height values for the body tag element
> 
> this was intentional because not all browsers supported the `dvh` unit so `vh` is a fallback
> 
> that being said Chrome has now added support for `dvh` so maybe we should consider removing the fallback https://caniuse.com/mdn-css_types_length_viewport_percentage_units_dynamic

ahhhhh i see okay then this is good to go

PR #245, opened 2022-12-22
feat(client): ✨ add service worker

nstringham:
This PR adds a service worker that does 2 things.
- caches a bunch of files to make the app load faster for returning users
- add a custom offline page (this is a requirement for the "add to home screen" feature on Android


gchamb:
how do I test this? Is the custom offline screen only for android?

nstringham:
> how do I test this? Is the custom offline screen only for android?

You can test it by emulating a loss of internet connection in dev tools or by shutting down the dev server and reloading the page.

In theory the offline screen would display on all browsers however iOS shows custom error messages when offline instead of using the offline page provided by the website. I’ve tested it in Chrome and Firefox on Android and Windows and in Safari on iOS.

gchamb (approved):
okay uh ha ha ik what's going on now...like 85% know whats going on

PR #264, opened 2022-12-31
feat: ✨ Catfishes know about other catfishes

### LINKED ISSUE #260: Catfishes know other catfish
On the role screen, the catfish should see the names of their catfish buddies. In the voting and chatting phase, catfish names are red to other catfish.

- [x] Catfishes should know each other
- [x] Catfishes name are red so they don't accidentally vote each other out 
- [x] Catfishes name are red in chat so you know you are talking to your catfish buddy

gchamb:
- catfishes know about each other when roles are displayed
- highlight name in red in matched name if they're both catfishes
- highlight name in red for other catfishes in vote if they're a catfish


client/src/components/ReceiveRole.svelte
```diff
@@ -49,7 +69,7 @@
     color: #1976d2;
   }
 
-  .catfish {
+  :global(.catfish) {
```
nstringham: are you sure this should be global?

client/src/components/Vote.svelte
```diff
@@ -60,7 +61,9 @@
           on:click={() => vote(lobby.uids[i])}
         >
           <AvatarImg {avatar} />
-          <span class="mdc-typography--subtitle1">{displayName ?? ""}</span>
+          <span class="mdc-typography--subtitle1 {catfishes?.includes(lobby.uids[i]) ? 'catfish' : ''}"
```
nstringham: this should use `class:` like this
https://github.com/cat-acronym-team/cat-me-if-you-can/blob/feba43c513a5c303f605f707598be9a2786fcaf3/client/src/components/ChatMessages.svelte#L69-L71

firestore.rules
```diff
@@ -18,7 +18,17 @@ service cloud.firestore {
...
+        function isCatfish(){
+          return get(/databases/$(database)/documents/lobbies/$(code)/privatePlayers/$(request.auth.uid)).data.role == "CATFISH";
+        }
+
+        function isCatfishResource(){
+          return resource.data.role == "CATFISH"
+        }
+
         allow get: if request.auth.uid == uid;
+        allow list: if isCatfish() && isCatfishResource();
```
nstringham, line 31: do we really want catfish to be able to read the entire privatePlayer document?

client/src/routes/game/+page.svelte
```diff
@@ -29,6 +29,11 @@
   let privatePlayer: PrivatePlayer | undefined = undefined;
   let unsubscribePrivatePlayer: Unsubscribe | undefined = undefined;
 
+  let privatePlayerCollection: CollectionReference<PrivatePlayer>;
+
+  let catfishes: string[] | undefined = undefined; // undefined for cats or spectators | string[] for catfishes
```
nstringham: why not an empty array?

client/src/routes/game/+page.svelte
```diff
@@ -156,6 +163,25 @@
       }
     }
   }
+
+  $: privatePlayer, getCatfishes();
+  function getCatfishes() {
```
nstringham, line 168: shouldn't we call unsubscribeCatfishes here?

nstringham (dismissed):

client/src/components/ReceiveRole.svelte
gchamb: I was using this class in different components so I thought so

firestore.rules
gchamb: They're only allowed to read other catfish private player docs in which only holds the their role(which we want them to know), stalker(wouldn't matter if they know because they're working together), and prompt in which they would have the same prompt.

client/src/routes/game/+page.svelte
gchamb: I think I understand but explain why?

client/src/components/ReceiveRole.svelte
nstringham: global does not load it globally it only applies it globally so that would only be guaranteed to work if ReceiveRole is always loaded at the same time as anywhere else that uses it

firestore.rules
nstringham: hmm I guess so but it feels wrong

client/src/routes/game/+page.svelte
nstringham: I'm pretty sure the way the code is currently written we resubscribe every time the private player changes (each time there is a new prompt for example) but we never unsubscribe so that will put a lot of unwanted stress on the database.

firestore.rules
gchamb: If you have a better idea lmk. We'd be only duplicating a little bit of data if you do have one

client/src/components/ReceiveRole.svelte
```diff
@@ -26,13 +30,29 @@
     },
   };
 
+  function getCatfishNames() {
+    if (catfishes == undefined) {
+      return;
+    }
```
nstringham: is this if still needed?

client/src/components/Vote.svelte
```diff
@@ -60,7 +61,9 @@
           on:click={() => vote(lobby.uids[i])}
         >
           <AvatarImg {avatar} />
-          <span class="mdc-typography--subtitle1">{displayName ?? ""}</span>
+          <span class="mdc-typography--subtitle1 " class:catfish={catfishes.includes(lobby.uids[i])}
+            >{displayName ?? ""}</span
```
nstringham: personally I like this
```html
<span>
  text
</span>
```

more than this
```html
<span
>text</span
>
```

client/src/components/Vote.svelte
```diff
@@ -60,7 +61,9 @@
           on:click={() => vote(lobby.uids[i])}
         >
           <AvatarImg {avatar} />
-          <span class="mdc-typography--subtitle1">{displayName ?? ""}</span>
+          <span class="mdc-typography--subtitle1 " class:catfish={catfishes.includes(lobby.uids[i])}
```
nstringham: you acedentaly added space in the `class`

client/src/routes/game/+page.svelte
nstringham: why is `unsubscribeCatfishes` not called if the state is ROLE?

client/src/theme/dark/_smui-theme.scss
```diff
@@ -33,3 +33,7 @@ a:visited {
 .error {
   color: theme.$error;
 }
+
+.catfish {
+  color: #d32f2f;
+}
```
nstringham: this code is duplicated in `ReceiveRole` maybe it would be best to move all three role colors into the theme

nstringham (dismissed):

client/src/routes/game/+page.svelte
gchamb: So you saying subscribe for that moment then unsubscribe? And a side note I think the getDocs function might be better since we're not using the newly updated documents to display to the user or anything

nstringham: No I'm saying unsubscribe before you subscribe

nstringham: Yeah getDocs makes sense to me

gchamb: > No I'm saying unsubscribe before you subscribe
Yeah, that makes sense.  You want to make sure you unsubscribe from the previous subscription before you subscribe again in the Role phases case.

PR #266, opened 2022-12-31
fix(client): :bug: service worker merge conflicts

nstringham:
This PR solves some merging issues that I somehow didn't notice in #245


gchamb (approved):
hopefully, they can solve the import CSS problem but the workaround seems to work

PR #307, opened 2023-01-16
feat(client): ✨ add google analytics

### LINKED ISSUE #301: Add Google Analytics
Connect Google Analytics so can we can keep an eye on traffic, clicks, etc.

gchamb:
- Initialized Google Analytics
- Log Sign Up Events for Sign Up Methods
- Log Lobby Creation
- Log Lobby Joining
- Log Games Starting
- Log Games Ending
- Log Account Deletions

If there is anything you think would be informative to know that I didn't log then let me know. I'll add it.

### DIFF
--- a/client/src/components/AccountButton.svelte
+++ b/client/src/components/AccountButton.svelte
@@ -12,6 +12,8 @@
   import type { UserData } from "$lib/firebase/firestore-types/users";
   import ProviderButtons from "./ProviderButtons.svelte";
   import PasswordTextfield from "./PasswordTextfield.svelte";
+  import { logEvent } from "firebase/analytics";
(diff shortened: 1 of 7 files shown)


nstringham:
These all seam like things that could be logged on the server side. Intuitively it seams like logging on the server side would be better because we don't need a bunch of network requests on the user's device.

What do you think?

gchamb:
> These all seam like things that could be logged on the server side. Intuitively it seams like logging on the server side would be better because we don't need a bunch of network requests on the user's device.
> 
> 
> 
> What do you think?

Yeah, we talked about it on the discord. Also, are there any other events we should log that would be meaningful to us?

